PR #11, opened 2018-04-26
WIP: Block System

Roger-luo:
### Properties
- `nqubits` number of qubits
- `cache_level` cache level of this block
- `full` get this block's dense matrix form
- `sparse` get this block's sparse matrix form

### Methods
- `apply!` apply this block to a register
- `update!` update parameters according to given sequence
- `cache!` force cache all blocks that under given cache level


Roger-luo:
## Refactor composite blocks

abstract type hierarchy:

```julia
abstract type AbstractBlock{N} end
abstract type LeafBlock{N} <: AbstractBlock{N} end
abstract type AbstractGate{N} <: LeafBlock{N} end
```

Do we still need a abstract block for unitaries?

```julia
abstract type UnitaryBlock{N} <: AbstractBlock{N} end
```

`Concentrator` is removed from composite blocks, composite blocks now only contains `ChainBlock`, `KronBlock`, these two are unitaries.

```julia
ChainBlock <: UnitaryBlock # or AbstractBlock
KronBlock <: UnitaryBlock # or AbstractBlock
```

### ChainBlock

`ChainBlock` now checks its contents' shape.

### KronBlock

Can `KronBlock` takes a `Concentrator` as its sub-block? I don't think so. `Concentrator` itself can not kroneckered with other unitaries. Therefore, a `UnitaryBlock` could be necessary for this type check.

### Measure

Measure will only be implemented as a method:

- `measure(reg, address)->bitstring` measure a result (a bit string), but do not change the original register state
- `measure!(reg, address)->bitstring, reg` measure a result (a bit string), and change the register state (to an eigen state)
- `measure_remove(reg, address) -> new_reg` measure a result and trace the qubits 

@GiggleLiu

Roger-luo:
Another change, since `Concentrator` is actually a rank-3 tensor, but the `full` & `sparse` methods are supposed to return a matrix (unitary) from block, it will now only return a subtype of `UnitaryBlock`'s matrix.

the matrix form of `ChainBlock` will be implemented to be able to generate unitary with correct size when it contains several `Concentrator` with correct shape.

Roger-luo:
```
const UnitaryBlock = Union{AbstractGate, ChainBlock, KronBlock}
```

Roger-luo:
## Cache Level

Add a new type parameter for each block types to specify its cache level. The block interface is extended to


```julia
apply!(block, reg, cache_level=0) # do not cache by default
cache!(block, level) # requires level, force update all the cache
dispatchin!(block, params) # dispatch params in block
```

Issue #16, opened 2018-04-28
[Register] Optimization for contiguous address (line orders)


Roger-luo:
By packing contiguous memory address together, `Concentrator` could be more efficient. (Less permutation operations on shape).

Currently, the mechanism is:

- if input is a `UnitRange`, this range of address will be processed as a contiguous quantum memory address.
- if input is a `Int` or `Vector{Int}`, this range of address will be processed as an in-contiguous quantum memory address.

We suppose that if a user inputs `UnitRange`, then it knows which part is contiguous, therefore there is no contiguous check for `UnitRange`, and we directly pack an address start from the start of `UnitRange` and has length `length(::UnitRange)`.

However, currently solution can not handle the following conditions:
[1, 6, 5, 4, 2, 3], where [6, 5, 4] is actually contiguous, but if the input is `4:6`, it will cause an error.

Roger-luo:
4, 5, 6 and 6, 5, 4 is different, we don't need to consider this anymore

PR #17, opened 2018-04-28
Register: Refactor

Roger-luo:
## Basics

- `AbstractRegister` is no more a subtype of `AbstractArray`. (array interface is removed)

## Default Register

- rename `ids` to `line_orders` (`address` would be better?)
- rename `data` to `state`
- `Register` stores the state as an `Array{T, 2}` (previously we allow arbitrary `Array{T, N}`)
- initialization is done by methods rather than constructors: `zero_state`, `rand_state`

### Packing and Focusing
- `pack!` methods only take one instance of `Register`.
- `pack!` is optimized for contiguous memory, but there is still problems #16 
- `focus!` now has side-effect, rename it from `focus`
- both `focus!` and `pack!` supports mixed range like `(2, 5, 6:7, 9, 10)`


Roger-luo:
### Exported APIs

- `nqubit`, `nbatch`, `line_orders`, `state`
- `eltype`, `copy`
- `focus!`
- `zero_state`, `rand_state`

Note: we don't export `Register` as an API, since user should use `zero_state` or `rand_state` to create a default register.

src/Register.jl
```diff
@@ -1,206 +1,147 @@
...
+but since it is not immutable (we need to change its shape),
+be careful not to change its state, though the behaviour is
+the same, but allocation should be avoided. Therefore, no
+shallow copy method is provided.
 """
-struct Register{M, B, T, N} <: AbstractRegister{M, B, T, N}
-    data::Array{T, N}
-    ids::Vector{Int}
+mutable struct Register{N, B, T} <: AbstractRegister{N, B, T}
+    state::Array{T, 2}
```
GiggleLiu: I made a mistake, now I think rank-3 tensor makes more sense, because in this way, the array shape contains all informations about qubit and batch. otherwise, we can not distinguish remaining dimension and batch dimension without query the register.

src/Register.jl
```diff
@@ -1,206 +1,147 @@
...
+    reg
 end
 
-function focus(src::Register{M, 1, T, M}, ids::NTuple{K, Int}) where {M, T, K}
-    pack!(src, ids)
-    exposed_size = 2^K
-    remained_size = 2^(M-K)
-    data = reshape(src.data, exposed_size, remained_size)
-    Register(M, 1, data, src.ids)
+exposed_size(orders::NTuple{K, Int}) where K = 2^K
```
GiggleLiu: nexposed_qubit, since you have nqubit API, it is more natural way to design API.
use 1<<n instead of 2^K! about 5 times faster!

PR #54, opened 2018-05-30
Intrinsics, boost performance

GiggleLiu:


GiggleLiu:
realize inverse control for `cxgate` et al.

@GiggleLiu

PR #101, opened 2018-06-23
fix and speedup log2i

JeffreySarnoff:
your implementation of log2i did not work -- no matter,  one arrow, two bullseyes

### DIFF
--- a/src/Intrinsics/Math.jl
+++ b/src/Intrinsics/Math.jl
@@ -9,17 +9,16 @@ bit_length(x::Int32)  =  32 - leading_zeros(x)
 """
     log2i(x::Integer) -> Integer
 
-Return log2(x), this integer version of `log2` is fast but only valid for number equal to 2^n.
-Ref: https://stackoverflow.com/questions/21442088
+Compute the base-2 logarithm of a nonnegative integer,
+rounding to zero.  Returns floor(Int, log2(x)) and
+is exact for nonnegative powers of 2.
+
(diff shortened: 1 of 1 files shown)


src/Intrinsics/Math.jl
```diff
@@ -9,17 +9,16 @@ bit_length(x::Int32)  =  32 - leading_zeros(x)
...
 """
-function log2i(x::T)::T where T
-    local n::T = 0
-    while x&0x1!=1
-        n += 1
-        x >>= 1
-    end
-    return n
-end
+log2i(x::Int64) = !signbit(x) ? (64 - leading_zeros(x)) : throw(ErrorException("nonnegative expected ($x)"))
```
Roger-luo: Thanks for your PR. It seems that this does not pass the test. I guess this should minus one? or this won't be equivalent to the definition of `log`. Besides, we may still need the original implementation as a fallback for other types. Or maybe just use `sizeof`

```Julia
log2i(x::T) where {T <: Integer} = !signbit(x) ? (8 * sizeof(x) - leading_zeros(x)) : throw(ErrorException("nonnegative expected ($x)"))
```

But not sure if this will make the total size of an integer (which should be a const) become some thing calculated in the runtime.

JeffreySarnoff: ok never mind

JeffreySarnoff: the original implementation returned zero for all inputs (or so it seemed)

PR #114, opened 2018-07-09
WIP: CUDA support

Roger-luo:
Though CUDA support is in the Roadmap. But we might have to wait for 
https://github.com/JuliaGPU/CuArrays.jl/pull/90

Then I will add a CUDA register and some basic block operations on GPU.


GiggleLiu:
Hooray! we have [CuYao.jl](https://github.com/QuantumBFS/CuYao.jl) now! Soon, we will have its 0.1.0 release.

PR #116, opened 2018-07-16
fix diff circuit

Roger-luo:
use put block instead, 6x faster than scipy implementation


Roger-luo:
@GiggleLiu Is this a `blockfilter` bug? `BlockTreeIterator` works fine.

Roger-luo:
This is fixed QuAlgorithmZoo (WIP), I'll close this.

PR #118, opened 2018-08-03
WIP: Fix 0.7 deprecations

Roger-luo:
- [x] drop v0.6 support in CI after v1.0 released


Roger-luo:
Need to wait for LuxurySparse.jl be registered in the METADATA

Issue #127, opened 2018-08-14
Move some examples to QuAlgorithmZoo.jl?


Roger-luo:
@GiggleLiu To avoid this kind of bug (#125 ) happen again and to keep a frequent update of various new algorithms, I think we can move the examples (maybe just some of them, or just all of them) to [QuAlgorithmZoo.jl](https://github.com/QuantumBFS/QuAlgorithmZoo.jl) and keep update that repo with new algorithms and demos in notebooks like QAOA and HHL and etc.

check their repo, it looks nice:

https://github.com/JuliaDiffEq
https://github.com/qojulia

They separated examples (or other instances) to different repo, then you can update and fix them more frequently without worrying about it is not in CI.

Roger-luo:
Okay, discussed with @GiggleLiu I will move examples to QuAlgorithmZoo.jl

PR #129, opened 2018-08-16
Move zoo

Roger-luo:
#127


Roger-luo:
I guess the appveyor failure is not our problem...

PR #130, opened 2018-08-18
Fix windows CI & update installation guide for Julia 1.0

Roger-luo:


Roger-luo:
I made a PR on the wrong branch #130

PR #135, opened 2018-08-31
Stability improvements

GiggleLiu:
* input type checks
* `focus!(reg, locs) do ... end` interface


Roger-luo:
I just merged your patch for LuxurySparse. Let's wait for the tag.

PR #138, opened 2018-10-02
Refactoring dipspatch! and parameter interfaces, new PauliString

GiggleLiu:
Autodiff will be added soon, in order to prepair for it, we refactor current parameter dispatch frameworks.

See issue #141 and #140

```julia
mutable struct Diff{N, T, GT<:Rotter{N, T}, RT<:AbstractRegister} <: NonParametricContainer{N, Complex{T}}
    block::GT
    output::RT
    grad::T
    Diff(block::Rotter{N, T}, output::RT) where {N, T, RT} = new{N, T, typeof(block), RT}(block, output, T(0))
    Diff(block::Rotter{N, T}) where {N, T} = Diff(block, zero_state(N))
end
```

The subtype tree
```julia
AbstractBlock
    AbstractMeasure
        Measure
        MeasureAndRemove
    FunctionBlock
    MatrixBlock
        AbstractContainer
            Concentrator
            ControlBlock
            Diff
            PutBlock
            RepeatedBlock
            TagBlock
                CachedBlock
                Daggered
                Scale
        CompositeBlock
            ChainBlock
            KronBlock
            PauliString
            Roller
        PrimitiveBlock
            ConstantGate
                CNOTGate
                HGate
                I2Gate
                P0Gate
                P1Gate
                PdGate
                PuGate
                ToffoliGate
                XGate
                YGate
                ZGate
            GeneralMatrixGate
            MathBlock
            PhaseGate
            ReflectBlock
            RotationGate
            ShiftGate
            Swap
            TimeEvolution
    Sequential
```


Roger-luo:
What's the difference between PauliString and a Clifford tableu?

GiggleLiu:
@Roger-luo 
`PauliString` is something like `XYIIZ`, so that we can represent a hamiltonian as a sumation of scaled PauliString.

It is `hermitian`, `unitary` and `reflexive`, so that can also be used in `Rotor`.

Maybe in the future, we will give a PR to `Clifford.jl` and use it as a backend. But now, `Clifford.jl` is not strong enough to represent a hamiltonian easily.

src/Blocks/IOSyntax.jl
```diff
@@ -85,7 +83,7 @@ print_tree(tree, args...; kwargs...) = print_tree(STDOUT::IO, tree, args...; kwa
 print_subblocks(io::IO, tree, depth, charset, active_levels) = nothing
 
 function print_subblocks(io::IO, tree::CompositeBlock, depth, charset, active_levels)
-    c = blocks(tree)
+    c = subblocks(tree)
```
Roger-luo: what's difference between `blocks` & `subblocks`

test/Blocks/Core.jl
```diff
@@ -37,37 +37,37 @@ end
...
+        @test parameters(g[3]) ≈ [2]
 
     end
 
     @testset "check function dispatch" begin
         g = chain(phase(0.0), phase(0.0))
         dispatch!(+, g, [1, 2])
-        @test parameters(g[1]) ≈ 1
-        @test parameters(g[2]) ≈ 2
+        @test parameters(g[1]) ≈ [1]
```
Roger-luo: `parameters` returns an iterator not a vector.

src/Blocks/IOSyntax.jl
GiggleLiu: `subblocks` is a renaming of `blocks`, since this name is confusing sometime, `sub` emphasis the tree depth is increased.
Every block has `sublocks` interface, which makes depth first search of tree much easier.

test/Blocks/Core.jl
GiggleLiu: `parameters` now always return a `Vector`, see the implementation in https://github.com/QuantumBFS/Yao.jl/issues/140 .

For iterable, you probably refers to `iparameters`, which is intrinsic parameters of a block.

src/Blocks/Container.jl
```diff
@@ -0,0 +1,43 @@
...
+get an iterator that iterate through all sub-blocks.
+"""
+function block end
+
+"""
+    chblock(block, blk)
+
+set the block of a container.
+"""
+function chblock end
```
Roger-luo: Why not just `setblock`?

src/Blocks/Scale.jl
```diff
@@ -0,0 +1,97 @@
+export Scale, Pos, Neg, Im, _Im, scale, getscale
+"""
+    Scale{X, N, T, BT} <: TagBlock{N, T}
+
+    Scale{X}(blk::MatrixBlock)
+    Scale{X, N, T, BT}(blk::MatrixBlock)
+
+Scale Block, by a factor of X, notice X is static!
```
Roger-luo: What's the different if this is not static? It's not clear to me there's  performance benefits by convert a floating point to static. You didn't dispatch specialized method for it (only except im).

1. Could you explain why we need this, rather than a simple runtime scaling?
2. Could you show a benchmark for this?

src/Blocks/TimeEvolution.jl
```diff
@@ -32,14 +32,10 @@ adjoint(blk::TimeEvolution) = TimeEvolution(blk.H, -blk.t)
 
 copy(te::TimeEvolution) = TimeEvolution(te.H, te.t)
 
-function dispatch!(te::TimeEvolution, itr)
```
Roger-luo: I actually know why you want this `iparameters` now. But I still don't think we actually need it, this add another layer of the abstraction (the first layer is the `dispatch!` and `parameters`)

Because your `setiparameters` and `niparameters` stays the same with the original (it's just a name change) It's possible to get the same result and also performance with only three methods `dispatch!`, `parameters`, `nparameters`.

But we can always fix this later. I'll keep this for now (I really don't have time to refactor the dispatch this month).

src/Interfaces/DeprecationWarnings.jl
```diff
@@ -0,0 +1,7 @@
+using Base: depwarn
+export blocks
+
+function blocks(blk::AbstractBlock)
```
Roger-luo: You should use `Base.@deprecate` rather than manually throw this warning.

Roger-luo: And a more standard way is to move all deprecations in a file called `deprecations.jl` in each module (so this should appear in `Yao.Blocks`), or if someone directly use `Yao.Blocks`, there won't be this warning at all.

src/Registers/Default.jl
```diff
@@ -7,39 +7,43 @@ export DefaultRegister
...
-mutable struct DefaultRegister{B, T} <: AbstractRegister{B, T}
-    state::Matrix{T} # this stores a batched state
+mutable struct DefaultRegister{B, T, MT<:AbstractMatrix{T}} <: AbstractRegister{B, T}
+    state::MT # this stores a batched state
 
-    function DefaultRegister{B, T}(raw::Matrix{T}) where {B, T}
-        ispow2(size(raw, 1)) && ispow2(size(raw, 2) ÷ B) ||
-            throw(InexactError(:DefaultRegister, DefaultRegister, raw))
-        new{B, T}(raw)
+    function DefaultRegister{B, T, MT}(raw::MT) where {B, T, MT<:AbstractMatrix{T}}
```
Roger-luo: Is this just the default constructor? remove this and move other constructor outside the struct?

src/Registers/Default.jl
```diff
@@ -7,39 +7,43 @@ export DefaultRegister
...
+register(raw::AbstractVector) = register(reshape(raw, :, 1))
 
 # Required Properties
 
 nqubits(r::DefaultRegister{B}) where B = log2i(length(r.state) ÷ B)
 nactive(r::DefaultRegister) = state(r) |> nqubits
 state(r::DefaultRegister) = r.state
-statevec(r::DefaultRegister{B}) where B = reshape(r.state, :, B)
-statevec(r::DefaultRegister{1}) = vec(r.state)
+relaxedvec(r::DefaultRegister{B}) where B = reshape(r.state, :, B)
```
Roger-luo: Can we add some simple comments or documents for `relaxedvec`?

src/Registers/Measure.jl
```diff
@@ -40,19 +40,24 @@ function measure_remove!(reg::AbstractRegister{B}) where B
         @inbounds res[ib] = ires
     end
     reg.state = reshape(nstate,1,:)
-    reg, res
```
Roger-luo: explain why the return value is changed?

src/Interfaces/DeprecationWarnings.jl
Roger-luo: But I don't think we need deprecations at all (or you should add all other changes to deprecation) since we are still in a beta version.

Roger-luo: Nah, I think we should just write some docs about what's different between v0.1 and v0.3 (before we release)

src/Blocks/Container.jl
GiggleLiu: `set` usually means inplace change, but here, we can not change block without side effect, since sub-block type are often used in type parameters.

src/Blocks/Scale.jl
GiggleLiu: `Scale{X}` is designed for `PauliGroup`, so that `X, -X, iX, -iX` can be dispatched as different types.
As you mensioned, we probably also need runtime scaling.

src/Blocks/Container.jl
Roger-luo: Then you should write change the block of a container in document.

src/Blocks/TimeEvolution.jl
GiggleLiu: > Because your setiparameters and niparameters stays the same with the original (it's just a name change)

They are different, we need both
* `iparameters` -> parameters declared in this block, what you can set directly inside this instance.
* `parameters` -> parameters in this `subtree`, what you can not set directly.

Only for `PrimitiveBlock`, they are same.

`niparameters` can always be regarded as a trait of block, but `nparameters` can not.

Roger-luo: but you don't have `niparameters` for `CompositeBlock` right? then this is not a trait for `CompositeBlock`

src/Interfaces/DeprecationWarnings.jl
GiggleLiu: OK,
I will just remove this depwarn, since this change has already been recorded in `NEWS.md`.

src/Blocks/TimeEvolution.jl
GiggleLiu: It is, it counts `iparameters`. For ANY block, `iparameters` returns `()` if not specified.
Our composite blocks do not contain `iparameters` so we don't need to overide them.
Now we can make `RotationGate` subtype `CompositeBlock` if we want, then it will be a composite block with parameters.

Roger-luo: I see, then what about overloading the dot operator? by overloading `Base.getproperty`, this is a new feature of Julia 1.0 and is widely used in libs like `DataFrames` now. 

So the syntax will looks like

```julia
gate = SomeGateType()
gate.parameters # this returns the iparameters, which is the parameters of this instance
parameters(gate) # this returns all the parameters
```

src/Registers/Measure.jl
Roger-luo: Agreed.

src/Blocks/Scale.jl
Roger-luo: Let's keep this first.

For Pauli String, I think you should be able to use Aaronson's tableau algorithm to optimize the calculation. And simplify the Pauli string. This should be some kind of optimizer rather than a block though.

GiggleLiu: Can you give an example for ‘optimizer’?

I agree that we need some arithsmatics to simplify PauliString, we can make another pr latter. 
PauliString can be used to represent Hamiltonian in time evolution. With better arithmetics, it will be convinient for Hamiltonian/MPO construction.

src/Blocks/TimeEvolution.jl
GiggleLiu: Let’s checkout its performance first.

Not very happy with overloading getptoperty, although I have been a strong supporter of this feature. It is not necessary and not Julia.

Roger-luo: Whose performance? You mean overloading `getproperty`'s performance? It's the same as a function. only syntax difference, because the parameters of an instance itself is its own property indeed, which make sense for overloading `getproperty`. 

(So nobody will be confused what's `iparameters` and `parameters`), it's easy to distinguish what are they made for.

src/Blocks/Scale.jl
Roger-luo: But in my mind, even we have `MPO` it will be in another package with its own `PauliString`. What I see here is that this block is actually an optimizer, it auto merge a `PauliString`.

An optimizer, is something that is able to find equivalent but simpler **block tree**, which is called optimizing the computation (by using less gates/gaining better performance).

Roger-luo: We'll just use `PauliString` for now, and expand it to a tableau in the future. #145

GiggleLiu: OK, that sound good. We can implement MPO in another package like [FunnyTN](https://github.com/QuantumBFS/FunnyTN.jl), the dependence hierarchy looks like
`Yao` depends on `MPO` depends on `CliffordAlgebra`

We should probabily respect [Cliffords.jl](https://github.com/BBN-Q/Cliffords.jl) and help improve this package? This package is simple but regularly maintained.

src/Blocks/TimeEvolution.jl
GiggleLiu: Yes, `getproperty` uses `if` and `else`, I don't know whether it will be optimized out.

Using `getproperty` can be a good idea, but let's put this problem aside at this moment. Since not only `iparameters`, but also `niparameters`, `subblocks`, `block`, `state` can be implemented using `getproperty`. For consistency consideration, it is should not be an easy decision.

Roger-luo: I'll file an issue for this. We support those in Julia 1.0 now, (we can't before).

Issue #137, opened 2018-09-27
Documenter is broken!


Roger-luo:
Just noticed.

Roger-luo:
Fixed. #142

PR #142, opened 2018-10-04
Adapt new workflow

Roger-luo:
In Julia 1.0, `REQUIRE` is deprecated, and with `Project.toml` you can

#### load this package locally

Clone the repo to anywhere you like

```sh
git clone git@github.com:QuantumBFS/Yao.jl.git Yao
cd Yao
```

start REPL with `julia --project`

image

and you can add `--project` in **Additional Julia Startup Arguments** in atom, if you use **Juno** for development

image

#### develop everything as you do in other languages

You can directly run a file in CLI as

```julia
julia --project xxx.jl
```

#### New Documenter configuration

Use new documenter configuration:  https://discourse.julialang.org/t/psa-use-a-project-for-building-your-docs/14974

## NOTE

I would personally use an `alias` for `julia --project`, and CLI tools for Julia is coming too.


Roger-luo:
`IterTools` and `Lazy` are deleted, due to:

- we have builtin `Iterators` now
- `@forward` is already in `MacroTools`

Roger-luo:
@GiggleLiu pls check if this works for you locally before you merge it (and this should fix the Documenter problem #137 as well).

.gitignore
```diff
@@ -14,4 +14,3 @@ _*.dat
 __pycache__/
 
 Manifest.toml
```
GiggleLiu: I need some folder like _local or _deprecated to store local trash like temporary notebooks.

Roger-luo: OK, we can keep it. add `_local` back

PR #149, opened 2018-10-11
Differentiation for V-statistics

GiggleLiu:
* bug fix
* opdiff
* vstatdiff
```julia
opdiff(psifunc, diffblock::AbstractDiff, op::MatrixBlock)
vstatdiff(psifunc, diffblock::AbstractDiff, vstat::Vstat; p0)
```

going to optimize QCBM doc!


GiggleLiu:
We should examplify autodiff in Yao, but I agree that we should use a simpler example.
After that, we can move it into QuAlgorithmZoo.

PR #152, opened 2018-10-22
Hamiltonian Representation

GiggleLiu:


src/Blocks/Container.jl
```diff
@@ -31,6 +31,20 @@ change the block of a container.
 """
 function chblock end
 
+"""
+    istraitkeeper(block) -> Bool
+
+change the block of a container.
+"""
+function istraitkeeper end
```
Roger-luo: What if someone define

```
istraitkeeper(::XXXBlock) = Val(true)
isunitary(x::XXXBlock) = (#=do something=#)
```

How do you solve the conflict?

Roger-luo: I'll merge this first, but this could remain to be a problem.

PR #153, opened 2018-10-24
WIP: Support QASM

Roger-luo:


Roger-luo:
I want to add a `@code_qasm`, which is just the same with others: `@code_llvm`, `@code_cuda`, etc.

Roger-luo:
Closed since this is going to be in another package now.

PR #155, opened 2018-11-02
Minor changes to support CuYao.jl

GiggleLiu:


src/Intrinsics/elementary.jl
```diff
@@ -150,15 +150,15 @@ end
 @inline unrows!(state::AbstractVector, inds::AbstractVector, U::IMatrix) = state
 
 @inline function unrows!(state::AbstractVector, inds::AbstractVector, U::SDDiagonal)
-    @simd for i in 1:length(U.diag)
```
Roger-luo: explain why this is removed?

docs/src/tutorial/GHZ.md
```diff
@@ -78,7 +78,7 @@ statevec(r)
 We have a GHZ state here, try to measure the first qubit
 
 ```julia
-measure(r, 1000)
+measure(r, nshot=1000)
```
Roger-luo: use `;` for kwargs

```
measure(r; nshot=1000)
```

src/Intrinsics/elementary.jl
GiggleLiu: 1). Compiler is clever enough to automatically put @simd. Benchmarks show no slow down if I remove it.
2). It causes trouble when using it inside GPU kernel functions.

docs/src/tutorial/GHZ.md
GiggleLiu: Is there some reason to use ‘;’  instead of ‘,’? If we change it here, we should also change it everywhere else.

Roger-luo: it sometimes mixed with default values. `;` looks more explicit and it is the standard way to define keyword arguments. e.g

https://docs.julialang.org/en/v1/base/arrays/#Base.cat

We should polish the doc string style at some point in principle. But let's use it from here.

GiggleLiu: I will fix some of them in this PR. Polishing all the doc strings is not urgent right now, let's keep this in mind.

GiggleLiu: Maybe we can wait for the automatic formatting tools.

Roger-luo: I don't think [DocumentFormat.jl](https://github.com/ZacLN/DocumentFormat.jl) will be able to do this in near future...

PR #158, opened 2018-12-17
Fix measure

GiggleLiu:
list of changes
1. `measure!(op, reg, loc)`
direct measure of operators.
2. `insert_qubit!`
insert qubits into register
3. fix `measure(reg, loc)` behavior when reg is focused.

### DIFF
--- a/docs/src/man/blocks.md
+++ b/docs/src/man/blocks.md
@@ -16,13 +16,6 @@ a quantum circuit into serveral kinds of blocks. The uppermost abstract type for
 
 ## Composite Blocks
 
-### Roller
-
-[`Roller`](@ref) is a special pattern of quantum circuits. Usually is equivalent to a [`KronBlock`](@ref), but we can optimize
-the computation by rotate the tensor form of a quantum state and apply each small block on it each time.
-
-![Block-System](../assets/figures/roller.svg)
-
 ## Blocks
 
 ```@autodocs
--- a/src/Blocks/CachedBlock.jl
+++ b/src/Blocks/CachedBlock.jl
@@ -39,7 +39,8 @@ clear!(c::CachedBlock) = (clear!(c.server, c.block); c)
 # forward methods
 function mat(c::CachedBlock)
     if !iscached(c.server, c.block)
-        m = dropzeros!(mat(c.block))
+        #m = dropzeros!(mat(c.block); trim=1e-16)
+        m = mat(c.block)
         push!(c.server, m, c.block)
         return m
     end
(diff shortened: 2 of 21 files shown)


src/Blocks/block_operations.jl
```diff
@@ -103,10 +103,29 @@ function expect end
...
 
-expect(op::MatrixBlock, dm::DensityMatrix) = mapslices(x->sum(mat(op).*x)[], dm.state, dims=[1,2]) |> vec
+#expect(op::MatrixBlock, dm::DensityMatrix) = mapslices(x->sum(mat(op).*x)[], dm.state, dims=[1,2]) |> vec
 expect(op::MatrixBlock, dm::DensityMatrix{1}) = sum(mat(op).*dropdims(dm.state, dims=3))
+function expect(op::AbstractBlock, dm::DensityMatrix{B}) where B
+    mop = mat(op)
+    [tr(view(dm.state,:,:,i)*mop) for i=1:B]
+end
+
+expect(op::AbstractBlock, reg::AbstractRegister{1}) = reg'*apply!(copy(reg), op)
```
Roger-luo: What's `reg'` is it `ConjReg` ? I think we should rename this to `AdjointReg` which is more consistent to stdlib.

docs/src/man/blocks.md
```diff
@@ -16,13 +16,6 @@ a quantum circuit into serveral kinds of blocks. The uppermost abstract type for
 
 ## Composite Blocks
 
-### Roller
```
Roger-luo: Why roller is deleted here? We haven't move them outside yet.

src/Blocks/block_operations.jl
GiggleLiu: Yes, it is. +1 for this change.

docs/src/man/blocks.md
GiggleLiu: This is documentation, we should introduce most basic & important blocks first.

src/Blocks/CachedBlock.jl
```diff
@@ -39,7 +39,8 @@ clear!(c::CachedBlock) = (clear!(c.server, c.block); c)
 # forward methods
 function mat(c::CachedBlock)
     if !iscached(c.server, c.block)
-        m = dropzeros!(mat(c.block))
+        #m = dropzeros!(mat(c.block); trim=1e-16)
+        m = mat(c.block)
```
Roger-luo: Why `dropzeros!` is removed? Is this causing any bug? @GiggleLiu

Roger-luo:
@GiggleLiu I think this PR and ConjReg are in YaoArrayRegister now, double check this and we could close this PR.

I'm not sure what was the bug about measure, so I don't if this is fixed in YaoArrayRegister or not. Double check?

Roger-luo:
This is done in other packages.

PR #168, opened 2019-03-21
Update LICENSE.md

ViralBShah:
I think the bottom part is not required. 

You probably want to add to the license:
`Copyright 2019, Yao.jl contributors`
or list your names


Roger-luo:
Ah, haha, yes, thanks! I didn't notice that.

PR #177, opened 2019-04-18
update README

Roger-luo:
1. codecov is removed since we don't need it anymore for meta package
2. framework is updated to list of component packages


README.md
```diff
@@ -71,8 +69,13 @@ Yao is a framework that is about to have the following features:
   - As a white-box simulator, rather than using a black box, users will be aware of what their simulation are doing right through the interface.
   - **Hierarchical APIs** from **low abstraction quantum operators** to **highly abstract** circuit block objects.
 
-![](docs/src/assets/figures/framework.png)
-The whole framework is highly **modularized**, researchers can extend this framework for different purposes.
+## Architecture
+
+Yao is a meta package based on several component packages in order to provide a highly modularized architecture, researchers and developers can extend the framework with different component packages for different purposes with minimal effort. The component packages includes:
```
GiggleLiu: I think these words contains no information. One can use them in any package.

Roger-luo: This is basically the same with original...

PR #184, opened 2019-04-30
bug fix

cbwang2016:


Roger-luo:
fixed by https://github.com/QuantumBFS/YaoBlocks.jl/pull/31

PR #192, opened 2019-05-13
WIP: the Yao paper Julia Con 2019 Proceedings

Roger-luo:
we could just work on this branch

For reference: https://proceedings.juliacon.org/guide/authors/

due is July 7th


Roger-luo:
I'll close this since I just decided to submit to quantum journal. I'll open another one when we have the paper on arxiv.

PR #195, opened 2019-06-06
polish README

Roger-luo:
Polish the README to make it more native based on @ViralBShah's suggestions. Thanks BTW.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,24 +53,6 @@ For CUDA support, see [CuYao.jl](https://github.com/QuantumBFS/CuYao.jl).
 
 Some quantum algorithms are implemented with Yao in [QuAlgorithmZoo](https://github.com/QuantumBFS/QuAlgorithmZoo.jl).
 
-## Motivation
-Comparing with state of art quantum simulators, our package is inspired by variational quantum circuit optimization.
-Variational quantum optimization algorithms like quantum circuit Born machine ([QCBM](https://arxiv.org/abs/1804.04168)), quantum approximate optimization algorithm ([QAOA](http://arxiv.org/abs/1411.4028)), variational quantum eigensolver ([VQE](https://doi.org/10.1038/ncomms5213)) and quantum circuit learning ([QCL](http://arxiv.org/abs/1803.00745)) et. al. are promising killer apps on near term quantum devices.
-These algorithms require the flexibility to tune parameters and have well defined patterns such as "Arbitrary Rotation Block" and "CNOT Entangler".
-
-In Yao, we call these patterns "blocks". If we treat gate or gate pattern as "block", then the framework can
-
-* be flexible to dispatch parameters,
-* cache matrices of blocks to speed up future runs,
-* allow hierarchical design of quantum algorithms
-
-Thanks to Julia's duck type and multiple dispatch features, user can
-
-* easily **extend** the block system by overloading specific interfaces
-* quantum circuit blocks can be dispatched to some **special method** to improve the performance in specific case (e.g. customized repeat block of H gate).
-
-
-
 ## Features
 
 Yao is a framework with the following features:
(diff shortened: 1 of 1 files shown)


Roger-luo:
@GiggleLiu good to merge?

README.md
```diff
@@ -86,6 +68,18 @@ Yao is a framework with the following features:
...
 
+## Motivation
+
+The study of variational quantum algorithms is gaining popularity. Variational quantum optimization algorithms like quantum circuit Born machine ([QCBM](https://arxiv.org/abs/1804.04168)), quantum approximate optimization algorithm ([QAOA](http://arxiv.org/abs/1411.4028)), variational quantum eigensolver ([VQE](https://doi.org/10.1038/ncomms5213)) and quantum circuit learning ([QCL](http://arxiv.org/abs/1803.00745)) et. al. are promising for near term quantum devices.
+These algorithms require the flexibility to tune parameters and to leverage circuit patterns such as "arbitrary rotation block" and "CNOT entangler". In a departure from traditional simulators and frameworks, we have designed and developed a framework along with an Intermediate Representation (IR) to represent, simulate and manipulate quantum circuits.
+
+Our design enables:
+
+* Hierarchical design of quantum algorithms and greater abstraction for quantum circuits
+* Heterogeneous computing
```
GiggleLiu: What is exactly heterogeneous computing?

Roger-luo: heterogeneous means it can run on different hardware. 

The design of block IR is hardware free. 

I could add a link to wiki for this.

GiggleLiu:
It does not change much, I think?

This package is for physists, this is important. My suggestion is

* Put some graphs that should be interesting to all, like benchmarks.
* Using a concrete example to show the advantage of *easy to use*, *Hirachical design*, *IR* directly, rather than speaking it out. If a feature can not be shown as an example, it is not true to user.

Roger-luo:
This goal of this PR is to make it more native in English. Not to change the contents. I'm working on documentation for benchmarks, will add it in another PR.

I don't think it is good to put examples to README, it is painful to read a huge README. It is better to be short and straight forward. We can provide links to documentation and examples. A glance of the interface (like a 3 line QFT code) might be sufficient, but I don't want to make README huge, that should be in documentation and front page.

GiggleLiu:
> its not good to put examples to README

Can't agree, but agree with you that README should not be too long, and current README is already too long. Making it concise does not mean containing no example at all.

I can look into it later.

Then I think it is ok to merge.

README.md
GiggleLiu: Remove it if it requres wiki, it is not the key feature. Do you mean GPU backend? Then just say GPU support.

Roger-luo: It is not only about GPU support, it is about the why is the design hardware free, what does this hardware free capable of. Heterogeneous is conceptually a more powerful and precise description of hardware free.

Roger-luo:
I mean we can't put all the examples to README, this is not a good README. A glance of code with reasonable size, like an example of QFT is good, since it only has 3 lines. But anything larger should be in documentation or a front page instead of putting it in README.

README.md
Roger-luo: Simply say it is GPU support cannot cover other work in `YaoExperiment`, `YaoGraphs`, and other possible direction that other people are interested in (like real hardware support in Regetti). Thus heterogeneous is better, if you think this needs some description, we could add a link. But you can't get rid of this concept if we are still dealing with this arch, and this is a very common feature among many softwares.

This concept has been describe for decades, if people don't understand it, they should read wiki.

GiggleLiu: Most framework supports hardware... I mean this is not a unique feature. Also, YaoExperiment has only basic support.

Roger-luo: Yeah, but I mean even it is basic, we have it, and we also have RBNF for QASM. The design of quantum block IR itself is mature to support heterogeneous computing.

So what would you suggest here? a link? or?

GiggleLiu: So far, we do not have robust support to QASM. So let's just remove it?

Roger-luo: You misunderstand what I meant here. Heterogeneous computing is a feature for quantum block IR (note there is "our design enables" as the condition), it's not about QASM and other features of Yao.

The whole point of getting rid of `T` and `MatrixBlock` is to make it a better design for heterogeneous computing (or it can be hardware related). This feature means you can extend Yao easily to your desired hardware, which doesn't have to be supported by ourselves. 

Without this feature of quantum block IR, you can't extend Yao to QASM/CUDA/etc without dealing with the hardware abstraction (like the precision of blocks).

Even you think heterogeneous computing is a hard word to understand, we can change the language, but it is a very important feature of quantum block IR.

Roger-luo:
Let's merge this and revise things after the paper is done?

GiggleLiu:
ok

PR #262, opened 2020-11-21
Delete deprecations.jl

RDxR10:


Roger-luo:
umm, I'm not sure why this was not removed, but I think in the future we probably should just define deprecations in each component packages.

PR #269, opened 2020-12-31
Revert "Delete deprecations.jl"

GiggleLiu:
Reverts QuantumBFS/Yao.jl#262


GiggleLiu:
Notes:

1. In the previous PR, the tests do not pass.
2. We need this file to handle deprecations, even if it is empty.

PR #283, opened 2021-06-01
Better display of docs

VarLad:
That would be step one


VarLad:
@Roger-luo Kindly fill the details 
I would do that myself, but I've to keep working on it :sweat_smile:

Roger-luo:
so if I understand correctly this copy the documentation to the site, but this will only updates the doc when there is a push in this repo, but I'm thinking how can we actually trigger new documentation build when the components package gets a new version, since a large portion of the docs actually comes from the docstring of component packages, it seems DiffEqDocs is doing this by manually tagging new versions of the documentation package.

VarLad:
Ohhhh.... okay... if I understand correctly, 
you're asking me to make a CI, which triggers this CI when YaoBase or YaoBlocks are updated? That will trigger the CI in the QuantumBFS.github.io repo to build the site again

The entire plan being that
1. I copy the docs folder to the site
2. Make an "intermediate branch" where the built docs go, along with the not-yet-built Franklin site and the scripts to glue the header/banner to the docs (which the CI takes care of)
3. Finally build the site from this "intermediate branch" and deploy. The scripts mentioned in step 2 get converted into the docs via Franklin.

VarLad:
Although, @Roger-luo the biggest problem I'm facing would be

Screenshot from 2021-06-02 18-01-00

Screenshot from 2021-06-02 18-03-54


Building the docs locally using the `julia make.jl local` only gives me the latest version of the docs..... This behavior would remain the same when I build from quantumbfs.github.io

I do need the other versions too right? How do I get them?

Roger-luo:
> Building the docs locally using the julia make.jl local only gives me the latest version of the docs..... This behavior would remain the same when I build from quantumbfs.github.io

Yes, Documenter will need to be trigger on tagged branch to create a tagged version, so uploading the source file of the docs might not be the solution, but uploading the build results to another repo instead of the gh-pages branch would be the way to go.

Roger-luo:
> you're asking me to make a CI, which triggers this CI when YaoBase or YaoBlocks are updated? That will trigger the CI in the QuantumBFS.github.io repo to build the site again

this could work, but maybe also complicated, I think in general no one cares what version the components are exactly, but the main issue is to update the docstring when there
is an update in upstream packages (and usually this is gonna be a patch version, so it is unlikely triggered by CompatHelper), so what about

1. we make a repo (and a package) that is only for documentation and tutorial source files
2. the `[deps]` field of this package's `Project.toml` will be fixed to patch version
3. when there is a new version, something like a CompatHelper, we can use CompatHelper at the beginning, but should use something that replaces the versions to latest instead of insert new version, because in our case, we always want component packages to be the latest
4. then we create a tag of this document package along, to trigger the doc build and tutorial build, this would guarantee we get an update when there is new version of the components

I can experiment a bit with this idea first, since you probably don't have write permission to things, I'll post what I find later in this PR.

VarLad:
Okay
I'm guessing you mean YaoBlocks and YaoBase when you say upstream packages？ 

We could stick this same CI in them too. Whenever there's a push in components（other than compat-helper）, they push here to append some arbitrary text to a file here.（We can make a file here called ComponentUpdates.txt or something like that）

This CI triggers and pushes into the site repo which then builds and deploys the site！

VarLad:
> but uploading the build results to another repo instead of the gh-pages branch would be the way to go.

Okay, then you want to build here and not in the QuantumBFS.github.io repo. I got that wrong I guess
If so, this makes my work easier 🙂
Also, then this PR is pointless, I guess. I should add this job in the CI where the documenter is deploying and use https://github.com/marketplace/actions/push-directory-to-another-repository

I'll try dishing out a solution to all of it in a few hours....

VarLad:
@Roger-luo Okay! Better(?) look for docs https://varlad.github.io/Yao.jl/dev/man/blocks.html

If thats okay, Change the last line of make.jl to `deploydocs(repo = "github.com/QuantumBFS/QuantumBFS.github.io.git", target = "build", branch="docs")`

and merge

VarLad:
Please look into the permissions accordingly
That would be all the work from this repo. The franklin integration remains, which will be managed from the quantumbfs.github.io repo

Roger-luo:
The color of this is a bit strange, 

image

can we just use the indigo theme but use Yao's themed blue color?

VarLad:
Okay!

VarLad:
@Roger-luo Actually, what would you mean by the Indigo theme? :sweat_smile:

Roger-luo:
https://github.com/JuliaDiff/DocThemeIndigo.jl

VarLad:
@Roger-luo Check again https://varlad.github.io/Yao.jl/dev/man/base.html

VarLad:
Also, I feel weird for not knowing this theme beforehand... Indigo looks nice as hell!

VarLad:
@Roger-luo SO.... would you be deploying this to the site or the Yao repo?

VarLad:
@Roger-luo Any problems with this one?

docs/make.jl
```diff
@@ -107,12 +33,8 @@ makedocs(
...
         canonical = ("deploy" in ARGS) ? "https://docs.yaoquantum.org/" : nothing,
         assets = [
-            "assets/main.css",
-            asset("https://yaoquantum.org/assets/main.css"),
-            asset("http://yaoquantum.org/favicon.ico"),
-            asset(
-                "https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css",
-            ),
+            "assets/themes/indigo.css",
+            asset("https://raw.githubusercontent.com/QuantumBFS/QuantumBFS.github.io/master/_assets/favicon-dark.ico", class=:ico),
```
Roger-luo, line 37: for maintainance reason, let's use the domain name version

https://yaoquantum.org/assets/logo-light.png

docs/make.jl
```diff
@@ -122,4 +44,4 @@ makedocs(
     pages = PAGES,
 )
 
-deploydocs(repo = "github.com/QuantumBFS/Yao.jl.git", target = "build")
+deploydocs(repo = "github.com/VarLad/Yao.jl.git", target = "build")
```
Roger-luo, line 47: you need to change this back

docs/src/assets/themes/indigo.css
```diff
@@ -0,0 +1,73 @@
+@font-face {
```
Roger-luo, line 1: don't upload the css, put it into the `.gitignore`

VarLad:
@Roger-luo Both the PRs are ready to merge!

Roger-luo:
you still need to remove the CSS from this PR

Roger-luo:
the indigo package will generate the CSS from SCSS, we don't want to store a temperory compile result here.

docs/make.jl
```diff
@@ -7,82 +7,9 @@ using YaoBlocks: Optimise
...
-                    HTMLWriter.render_settings(ctx),
-                ),
-            ),
-            scripts...,
-        ),
-    )
-end
+#Venerable Inventor :)
 
+download("yaoquantum.org/assets/logo-light.png", "docs/src/assets/logo.png")
```
Roger-luo, line 12: why do we download the logo here?

VarLad:
> for maintainance reason, let's use the domain name version
> yaoquantum.org/assets/logo-light.png

Because you mentioned so?
Or I seem to be misunderstanding something...?

VarLad:
Also, @Roger-luo DocThemeIndigo doesn't have Yao's blue overalls
I got that theme from my own fork of DocThemeIndigo https://github.com/VarLad/DocThemeIndigo.jl

VarLad:
I removed the logo from assets and download it from the yaoquantum link

Roger-luo:
> I removed the logo from assets and download it from the yaoquantum link

you can use the link directly instead of downloading the logo yourself.

> Also, @Roger-luo DocThemeIndigo doesn't have Yao's blue overalls
> I got that theme from my own fork of DocThemeIndigo https://github.com/VarLad/DocThemeIndigo.jl

Ah right, I forgot that, this looks good now then.

VarLad:
> you can use the link directly instead of downloading the logo yourself.

Isn't that what I did? Using `download()`?
If you're telling me to use `assets("yaoquantum.org/assets/logo-light.png")`, I'm not sure what would happen to the dark mode of the theme, we need to save it as `logo.png` to apply it to both modes, and `assets` doesn't have a rename option.

Roger-luo:
OK to my suprise Documenter doesn't support asset rename or custom logo name. this looks good now.

PR #284, opened 2021-06-08
Injecting the ToC into Yao docs

VarLad:


Roger-luo:
This looks good in general, but the layout is a bit strange - can we have both the sidebar and Toc centerized instead of only centerize the content?

VarLad:
Isn't it already centralized?
I just removed the numbers and the scroll line though

Look here again https://varlad.github.io/Yao.jl/dev/man/base.html

VarLad:
Also, sorry for the late response, strange that Github didn't notify me.... :sweat_smile:

Roger-luo:
Yeah this does look better, I think I'll just merge this for now. The right TOC bar still looks a bit strange to me, perhaps it's because of the position. But we can improve this later.

VarLad:
I think it looks strange because the ToC at the left is not centralized

So.... maybe we should place the ToC in the right to far right too?

Screenshot from 2021-06-29 13-45-07

VarLad:
Opinions are welcome
Github is not notifying me about any of this so thats strange....

Again, sorry for the late response :sweat_smile:

VarLad:
........Google was marking all my github notifications as Spam.... :rofl:

PR #314, opened 2021-12-08
move QuAlgorithmZoo to lib

Roger-luo:
tutorial depends on this QuAlgorithmZoo, currently dev this in CI slow down the build time. I think we can just move it into `lib` first, and improve them later


GiggleLiu:
I'd prefer delay this decision. Why are the tests rely on QuAlgorithmZoo?

Roger-luo:
not tests, documentation depends on this package

Roger-luo:
close in favor of #315

PR #315, opened 2021-12-09
move some utilities in YaoExtensions to Yao.EasyBuild submodule.

GiggleLiu:
Some blocks and circuit builders.

## List of blocks
* [ ] ReflectGate
* [ ] Bag
* [ ] EchoBlock
* [x] FSimGate
* [ ] ConditionBlock
* [ ] PauliString
* [ ] RotBasis
* [ ] MathGate
* [ ] Mod
* [ ] QFTBlock

## List of shortcuts
* [x] `cphase` and `CPhaseGate`
* [x] singlet_block
* [x] ISWAP
* [x] SqrtX
* [x] SqrtY
* [x] SqrtW
* [x] eyeblock
* [x] fsim_gate

## List of circuits
* [x] general_U4
* [x] phase_estimation_circuit
* [x] qft_circuit
* [x] rand_google53
* [x] hadamard_test_circuit
* [x] swap_test_circuit
* [x] rand_supremacy2d
* [x] variational_circuit

## List of hamiltonians
* [x] heisenberg
* [x] transverse_ising

NOTE 1: they are not exported to the global scope, one has to type
```julia
using Yao.EasyBuild
```
to access these features in Yao's submodule.

NOTE 2: the reason why some blocks are not added
* Mod: too specialized for Shor's algorithm, better place it in QuAlgorithmZoo.
* QFTBlock: requires FFTW, I do not want to add one more dependency for it.
* MathGate: this block is hard to use correctly.

It does not mean other blocks are useful, will consider removing them in the future, they are in the deleting list. When a block can be replace by a combination of other blocks, I prefer not creating a new type except there is a practical using case.


src/EasyBuild/block_extension/Bag.jl
```diff
@@ -0,0 +1,53 @@
+export AbstractBag
```
Roger-luo, line 1: what's the use case for `Bag` this looks like a complicated type to me, besides it has an abstract type

GiggleLiu: You are right, I removed some other blocks too. What about now?

src/EasyBuild/general_U4.jl
```diff
@@ -0,0 +1,47 @@
...
+
+export general_U2, general_U4
+
+"""
+    general_U2(θ1, θ2, θ3; ϕ=nothing)
+
+A general single qubits gate: ``e^(iϕ)R_z(θ_3)R_y(θ_2)R_z(θ_1)``.
+Leave `ϕ` as `nothing` to fix the global phase.
+"""
+function general_U2(θ1, θ2, θ3; ϕ=nothing)
```
Roger-luo, line 13: should this just be a block? this is also needed to map to openQASM etc.

Roger-luo (approved):
other parts LGTM

src/EasyBuild/general_U4.jl
GiggleLiu: It sounds good, but let us do it in another PR. Let us convert your comment to an issue?

PR #320, opened 2021-12-20
add n operator

Roger-luo:
moved from https://github.com/QuantumBFS/YaoBlocks.jl/pull/176/files


Roger-luo:
define this in EaRyd instead, we don't want to have `N` here given it's for Rydberg specifically

PR #328, opened 2022-01-23
Docs fixes

achuchmala:
* fix in-line math displaying according to suggestion in https://discourse.julialang.org/t/problem-with-inline-math-for-literate-generated-notebook/60132
* remove unnecessary lines
* fix link to Wikipedia article
* remove section about concentrator (deprecated)

### DIFF
--- a/docs/src/quick-start/2.qft-phase-estimation/main.jl
+++ b/docs/src/quick-start/2.qft-phase-estimation/main.jl
@@ -121,8 +121,6 @@ QFT(5)'
 
 # ![phase estimation](assets/phaseest.png)
 
-using Yao
-
 # First we call Hadamard gates repeatly on first `n` qubits.
 
 Hadamards(n) = repeat(H, 1:n)
@@ -131,7 +129,7 @@ Hadamards(n) = repeat(H, 1:n)
 
 ControlU(n, m, U) = chain(n+m, control(k, n+1:n+m=>matblock(U^(2^(k-1)))) for k in 1:n)
 
-# each of them is a `U` of power ``2^(k-1)``.
+# each of them is a `U` of power ``2^{(k-1)}``.
 
 # Since we will only apply the qft and Hadamard on first `n` qubits,
 # we could use `Subroutine`, which creates a context of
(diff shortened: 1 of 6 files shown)


docs/src/quick-start/2.qft-phase-estimation/main.jl
```diff
@@ -147,38 +145,15 @@ PE(n, m, U) =
 # other `m` qubits as the input state which corresponds to an eigenvector of
 # oracle matrix `U`.
 
-# The concentrator here uses `focus!` and `relax!` to manage
-# a local scope of quantum circuit, and only active the first `n` qubits while applying
```
Roger-luo, line 151: Why these parts are deleted?

docs/src/quick-start/2.qft-phase-estimation/main.jl
```diff
@@ -121,8 +121,6 @@ QFT(5)'
 
 # ![phase estimation](assets/phaseest.png)
 
-using Yao
```
Roger-luo, line 124: I think `using Yao` is also necessary?

Roger-luo (changes requested):
thanks! I left a few small comments

docs/src/quick-start/2.qft-phase-estimation/main.jl
achuchmala: OK, maybe I was too fast with removing this section :/. I've done this because we don't have `concentrator` here, which is now deprecated and replaced with `subroutine`.  However now I see that `subroutine` uses `focus!` and `relax!`, so here only change of `concentrator` to `subroutine` without deleting this description should be sufficient.

docs/src/quick-start/2.qft-phase-estimation/main.jl
achuchmala: Here in the QPE circuit we need QFT, which was defined above and needed `using Yao` anyway. If we have QFT and QPE in one file, then there is no need to repeat this line.

docs/src/quick-start/2.qft-phase-estimation/main.jl
Roger-luo: Yeah I see why you removed it, maybe just update the name? It's named `Subroutine` and `subroutine` now

Roger-luo: I think we still need a section to describe how `focus!` and `relax!` works

achuchmala: You are right, this description is valuable and gives some insight in how the subroutine works. Done :)

Roger-luo (approved):

Roger-luo:
Thanks you!

PR #334, opened 2022-02-12
remove batch dimension

GiggleLiu:
Since Julia can handle constants better, we do not need the batch size as the type parameters any more.


lib/YaoArrayRegister/src/register.jl
```diff
@@ -1,50 +1,140 @@
...
 !!! warning
 
-    `ArrayReg` constructor will not normalize the quantum state. If you need a
+    `BatchedArrayReg` constructor will not normalize the quantum state. If you need a
     normalized quantum state remember to use `normalize!(register)` on the register or
     normalize the input raw array with `normalize` or [`batched_normalize!`](@ref).
 """
-ArrayReg{B}(raw::AbstractMatrix; nlevel=2) where B = ArrayReg{B,nlevel}(raw)
-function ArrayReg{B,D}(raw::MT) where {B,D,T,MT<:AbstractMatrix{T}}
+mutable struct BatchedArrayReg{D,T,MT<:AbstractMatrix{T}} <: AbstractArrayReg{D,T,MT}
```
Roger-luo, line 63: we can let `BatchedArrayReg` be a subtype of `AbstractVector{ArrayReg}` then the rest iterative API will just make sense as a Vector and then remove the old `viewbatch` etc? given `BatchedArrayReg` is already a separate type. Maybe the name can be `ArrayRegBatch` instead which seems more intuitive

lib/YaoArrayRegister/src/register.jl
```diff
@@ -1,50 +1,140 @@
...
+function _warn_type(raw::AbstractMatrix{T}) where T
+    T <: Complex || @warn "Input matrix element type is not `Complex`, got `$(eltype(raw))`"
+end
+
+"""
+    arrayreg(state; nbatch::Union{Integer,NoBatch}=NoBatch(), nlevel::Integer=2)
+
+Create an array register, if nbatch is a integer, it will return a `BatchedArrayReg`.
+"""
+function arrayreg(state; nbatch::Union{Integer,NoBatch}=NoBatch(), nlevel::Integer=2)
```
Roger-luo, line 123: why not just use the constructor? This interface seems unintuitive to me, it's easier to just use the constructor of the corresponding type, then `NoBatch` won't be necessary.

Roger-luo (commented):
I think we should just change the concept of a register to be equivalent to a state then treat batched-register as optimization via dispatch would make more sense under this change. Otherwise, the definition of `BatchedArrayReg` seems not consistent with `ArrayReg` anymore.

lib/YaoArrayRegister/src/register.jl
GiggleLiu: It may also return you a BatchedArrayReg

lib/YaoArrayRegister/src/register.jl
GiggleLiu: All the previous methods are implemented for contiguous storage. This change is too big that might not be suited for this pr.

Roger-luo: there is no conflict here, it is still contiguous? the only change is

```julia
struct ArrayRegBatch{D, S <: AbstractMatrix} <: AbstractVector{ArrayReg{D}}
   storage::S
end
```

this will remove the batch concept from the core definition of registers

lib/YaoArrayRegister/src/operations.jl
```diff
@@ -9,25 +9,25 @@ export isnormalized, normalize!, regadd!, regsub!, regscale!, norm
 
 Check if the register is normalized.
 """
-isnormalized(r::ArrayReg) =
+isnormalized(r::AbstractArrayReg) =
```
Roger-luo, line 12: an example if we just subtype `AbstractVector{ArrayReg{D}}` these overloads won't be necessary for a `BatchedArrayReg` or `ArrayRegBatch`

lib/YaoArrayRegister/src/density_matrix.jl
```diff
@@ -29,59 +27,31 @@ function YaoBase.density_matrix(reg::ArrayReg, qubits)
     freg = focus!(copy(reg), qubits)
     return density_matrix(freg)
 end
-YaoBase.density_matrix(reg::ArrayReg{1}) = DensityMatrix(reg.state * reg.state')
-function YaoBase.density_matrix(reg::ArrayReg{B}) where {B}
```
Roger-luo, line 33: why this doesn't support batch anymore?

GiggleLiu: We do not have batched density matrix any more.

GiggleLiu: Because we do not have any batched function for it.

lib/YaoArrayRegister/src/register.jl
GiggleLiu: Let us keep an issue for it. I feel if we still want to apply a gate to it, it is intuitive to be a AbstractRegister.

lib/YaoArrayRegister/src/density_matrix.jl
Roger-luo: OK then if I already have a batched register how do you get for example entanglement entropy or other density matrices related quantities out from it? e.g use support broadcast on a batched register instead?

```julia
entropy.(batched_reg)
```

is this type of general broadcasting supported as the moment?

lib/YaoArrayRegister/src/register.jl
Roger-luo: yeah, I agree broadcasting `apply!` seems overcomplicated. But the current one is also not very intuitive when you do algebraic operations or other things on it. I could write up a PR for my proposal after this PR is merged. so let's not tag any new version before all these things are sorted out

GiggleLiu: Sounds good~

lib/YaoArrayRegister/src/density_matrix.jl
GiggleLiu: You mean this?
```julia
julia> using YaoArrayRegister

julia> reg = rand_state(6; nbatch=5)
BatchedArrayReg{2, ComplexF64, Transpose...}
    active qudits: 6/6
    nlevel: 2
    nbatch: 5

julia> von_neumann_entropy(reg, [2,3,4])
5-element Vector{Float64}:
 1.5744541948295039
 1.6045333650832052
 1.6710972251024965
 1.635335399305363
 1.5950027685238202
```

Roger-luo (approved):
LGTM

lib/YaoArrayRegister/src/density_matrix.jl
LWKJJONAK: Hi, I up Yao from v0.6.5 to v0.7.4 today and I found that the density_matrix cannot handle BatchedArrayReg now as you discussed above. I had to go back to v0.6.5 and everything works. Is this an unsolved issue?

GiggleLiu: We droped the batched density matrix support, because it makes things easier to maintain.
You can use `density_matrix.(reg)` instead to create a vector of density matrices.

PR #336, opened 2022-02-14
rm type parameter N

GiggleLiu:
Remove the type parameter `N` from the block system. (only keep this type parameter for `ConstGate`).
Lower case interfaces are not changed.


Roger-luo (approved):
LGTM

PR #346, opened 2022-03-15
fix empty chain adjoint

GiggleLiu:


GiggleLiu:
After a few rounds of testing, I am pretty sure the error in tests is from some bug in either Julia compiler or Test.

PR #350, opened 2022-03-23
fix 349 and 348

### LINKED ISSUE #348: Gradient of time_evolve
I'm implementing a custom function that is supposed to use QAOA ansatz, however, when I attempt to calculate the gradients I get an error about the dimension mismatch between the parameters and the gradient. Here is a minimal example that I could come up that reproduces this error.

```julia
using Yao, YaoExtensions
using Zygote

n = 2
h = transverse_ising(n)
U1 = variational_circuit(n, 1)

function wrap(U::AbstractBlock{N}, θ::Vector{Float64}) where N
    reg = apply(zero_state(n), dispatch(U, θ))
    real(expect(h, reg))
end

gradient(p -> wrap(U1, p), parameters(U1)) # this works

hb = sum([put(n, i => X) for i = 1:n])
U2 = chain(n, [time_evolve(hb, 0.0)])

wrap(U2, parameters(U2)) # this works
gradient(p -> wrap(U2, p), parameters(U2)) # this doesn't work
```

If I use an ansatz that is built out of basic gates like the one given by `variation_circuit`, then everything works out, however, when the ansatz contains `time_evolve` then I get the following error

```julia
ERROR: LoadError: DimensionMismatch("variable with size(x) == (1,) cannot have a gradient with size(dx) == (2,)")
Stacktrace:
 [1] (::ChainRulesCore.ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ChainRulesCore.ProjectTo{Float64, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}}}}})(dx::Vector{Float64})
   @ ChainRulesCore C:\Users\rdras\.julia\packages\ChainRulesCore\ksvfu\src\projection.jl:227
 [2] _project
   @ C:\Users\rdras\.julia\packages\Zygote\cCyLF\src\compiler\chainrules.jl:182 [inlined]
 [3] map(f::typeof(Zygote._project), t::Tuple{Vector{Float64}}, s::Tuple{Vector{Float64}})
   @ Base .\tuple.jl:246
 [4] gradient(f::Function, args::Vector{Float64})
   @ Zygote C:\Users\rdras\.julia\packages\Zygote\cCyLF\src\compiler\interface.jl:77
...
```

### LINKED ISSUE #349: chainrules_patch might be broken
```
cost = sum([put(N, i=>Z) for i in 1:N])
theta = rand(96)
N = 6
new_var = dispatch(EasyBuild.variational_circuit(N,5), :random);
function new_loss(theta)
	new_circ = dispatch(new_var, theta)
	diff_evalue = expect(cost, zero_state(N) => new_circ)
	return real.(diff_evalue)
end
```

Multiplying the return value to a constant (for example, `return real(2*diff_evalue)`) seems to have no effect on the calculated gradients.

GiggleLiu:
fix #349 fix #348

### DIFF
--- a/lib/YaoBlocks/src/autodiff/chainrules_patch.jl
+++ b/lib/YaoBlocks/src/autodiff/chainrules_patch.jl
@@ -18,7 +18,7 @@ end
 # primitive blocks
 unsafe_primitive_tangent(::Any) = NoTangent()
 unsafe_primitive_tangent(x::Number) = x
-for GT in [:RotationGate, :ShiftGate, :TimeEvolution, :PhaseGate]
+for GT in [:RotationGate, :ShiftGate, :PhaseGate]
     @eval function recursive_create_tangent(c::$GT)
         lst = map(fieldnames(typeof(c))) do fn
             fn => unsafe_primitive_tangent(getfield(c, fn))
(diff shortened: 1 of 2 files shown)


GiggleLiu:
Now this PR also fixes #348

PR #353, opened 2022-03-30
Enforce `Matrix{Basic}` type during conversion

wildart:
The `_pretty_basic` call mixes original element matrix type. By default `S` gate matrix is
```julia
julia> mat(S)
2×2 LinearAlgebra.Diagonal{ComplexF64, Vector{ComplexF64}}:
 1.0+0.0im      ⋅    
     ⋅      0.0+1.0im

```
but `mat(Bacis, S)` returns `Number`-typed matrix 
```julia
julia> mat(Bacis, S)
2×2 Matrix{Number}:
 1   0
 0  im
```
This makes impossible to use matrix multiplication.
```julia
julia> q
2×1 SparseArrays.SparseMatrixCSC{Basic, Int64} with 2 stored entries:
 (1/2)*sqrt(2)
 (1/2)*sqrt(2)

julia> mat(Bacis, S) * q.state
MethodError: no method matching zero(::Type{Any})
Closest candidates are:
  zero(::Type{Union{Missing, T}}) where T at ~/Applications/julia-1.7.2/share/julia/base/missing.jl:105
  zero(::Union{Type{P}, P}) where P<:Dates.Period at ~/Applications/julia-1.7.2/share/julia/stdlib/v1.7/Dates/src/periods.jl:53
  zero(::BitBasis.BitStr{N, T}) where {N, T} at ~/.julia/packages/BitBasis/8Z70U/src/bit_str.jl:51
  ...

Stacktrace:
 [1] zero(#unused#::Type{Any})
   @ Base ./missing.jl:106
 [2] mul!(C::Matrix{Any}, X::Matrix{Number}, A::SparseArrays.SparseMatrixCSC{Basic, Int64}, α::Bool, β::Bool)
   @ SparseArrays ~/Applications/julia-1.7.2/share/julia/stdlib/v1.7/SparseArrays/src/linalg.jl:95
 [3] *(X::Matrix{Number}, A::SparseArrays.SparseMatrixCSC{Basic, Int64})
   @ SparseArrays ~/Applications/julia-1.7.2/share/julia/stdlib/v1.7/SparseArrays/src/linalg.jl:115
 [4] top-level scope
   @ In[24]:1
 [5] eval
```

This PR correct type of the matrix
```julia
julia> mat(Basic, S)
2×2 Matrix{Basic}:
 1   0
 0  im

julia> mat(Basic, S)*q.state
2×1 Matrix{Basic}:
  (1/2)*sqrt(2)
 1/2*im*sqrt(2)
```

### DIFF
--- a/lib/YaoSym/src/symengine/register.jl
+++ b/lib/YaoSym/src/symengine/register.jl
@@ -22,7 +22,7 @@ function SymReg(r::AbstractArrayReg{D,<:Number}) where {D}
     return arrayreg(SparseMatrixCSC(smat); nbatch=nbatch(r), nlevel=D)
 end
 
-_pretty_basic(x) = x
+_pretty_basic(x) = Basic(x)
 _pretty_basic(x::Real) = isinteger(x) ? Basic(Int(x)) : Basic(x)
 function _pretty_basic(x::Complex)
     if isreal(x)
--- a/lib/YaoSym/src/symengine/register.jl
+++ b/lib/YaoSym/src/symengine/register.jl
@@ -23,7 +23,7 @@ function SymReg(r::AbstractArrayReg{D,<:Number}) where {D}
 end
 
 _pretty_basic(x) = x
-_pretty_basic(x::Real) = isinteger(x) ? Int(x) : x
+_pretty_basic(x::Real) = isinteger(x) ? Basic(Int(x)) : Basic(x)
 function _pretty_basic(x::Complex)
     if isreal(x)
         return _pretty_basic(real(x))
--- a/lib/YaoSym/test/symengine/blocks.jl
+++ b/lib/YaoSym/test/symengine/blocks.jl
@@ -20,8 +20,10 @@ end
 
 @testset "mat" begin
     @vars θ γ η
-    for G in [X, Y, Z, ConstGate.T, H, ConstGate.Tdag, ConstGate.T, I2, (0.5+0.5im)*X]
-        @test Matrix(mat(Basic, G)) ≈ Matrix(mat(G))
+    for G in [X, Y, Z, ConstGate.T, H, ConstGate.S, ConstGate.Sdag, ConstGate.Tdag, ConstGate.T, I2, (0.5+0.5im)*X]
+        B = mat(Basic, G)
+        @test eltype(B) === Basic
+        @test Matrix(B) ≈ Matrix(mat(G))
         @test Matrix(mat(Basic, control(4, 3, 2 => G))) ≈ Matrix(control(4, 3, 2 => G))
     end
 


Roger-luo:
good catch! Yeah, I agree the element type of returned matrix should be `Basic`, can we also have a test of this?

GiggleLiu:
The best way is to add a type annotation to the `_pretty_basic` function.
You fix will change the matrix type, which is not expected.

wildart:
I'm curious, what is the purpose of the `_pretty_basic` function? Why real integers converter to integer type? If a  prettified output is needed, it can be done in `show` function.

Roger-luo:
The problem is actually because `Complex` number's imaginary part is not recognized by SymEngine expression so one needs to manually convert a complex number and its imaginary unit to be a `Basic` if it's actually complex number and vice versa. Otherwise SymEngine will not be able to process the expression correctly. I think with a better symbolic system that has domain symbols like `SymComplex` and `SymReal` would solve the problem, but we don't really have such thing in Julia yet.

GiggleLiu:
Yeah, clean symbolic representation is very important in YaoSym.

wildart:
From what I see, SymEngine allows conversion of `Complex` numbers to `Basic` type: https://github.com/symengine/SymEngine.jl/blob/dc78e4ac4eafd353581cdfc6b710d59322d95b00/src/types.jl#L91

As well as Complex array conversion.

```julia
julia> using SymEngine

julia> a = [1 -3im 1+2im]
1×3 Matrix{Complex{Int64}}:
 1+0im  0-3im  1+2im

julia> b = convert(Matrix{Basic}, a)
1×3 Matrix{Basic}:
 1  -3*im  1 + 2*im

julia> a = [1 -3im 1.0+2im]
1×3 Matrix{ComplexF64}:
 1.0+0.0im  0.0-3.0im  1.0+2.0im

julia> b = convert(Matrix{Basic}, a)
1×3 Matrix{Basic}:
 1.0 + 0.0*im  -0.0 - 3.0*im  1.0 + 2.0*im
```

GiggleLiu:
In symbolic programming, you'd better get rid of floating point numbers. If you have floating point notation, the effect propagates to rotation gates. For example, are you happier with this?
```julia
julia> mat(Basic, X)
2×2 Matrix{Basic}:
            0  1.0 + 0.0*im
 1.0 + 0.0*im             0

julia> mat(Basic, rot(X, Basic(:θ)))
2×2 Matrix{Basic}:
                 cos((1/2)*θ)  (-0.0 - 1.0*im)*sin((1/2)*θ)
 (-0.0 - 1.0*im)*sin((1/2)*θ)                  cos((1/2)*θ)
```

Or this?
```julia
julia> mat(Basic, rot(X, Basic(:θ)))
2×2 SparseArrays.SparseMatrixCSC{Basic, Int64} with 4 stored entries:
     cos((1/2)*θ)  -im*sin((1/2)*θ)
 -im*sin((1/2)*θ)      cos((1/2)*θ)
```

We are not considering whether it is defined or not. The whole point of symbolic programming is make the output more like a mathematic notation.

Roger-luo:
Basic is not a Complex so you won't be able to say anything about the complex domain in Julia after using SymEngine expressions.   All you can see is a Number type. The issue is not about printing again but is about interop between Basic and Julia Complex. 

If you want to understand this problem better see my original proposal in Julia: https://github.com/JuliaLang/julia/issues/33246

We also had some old experimental symbolic system implemented in very early stage which demonstrates this problem: https://github.com/Roger-luo/Sym.jl

The real issue is to have a proper domain system like what people do in Mathematica or SymbolicsUtils and the domain system needs to work with Julia type system so that things can make sense.

GiggleLiu:
Well, it is a design choice: whether a symbol should carry a type or not. When you say the type information is important, you need to be more concrete about why "interop between Basic and Julia Complex" is an issue.

By not using floating point number, I mean they are not symbols. Another example is the `H` gate which we handled manually.

```julia
julia> mat(Basic, H)
2×2 Matrix{Basic}:
 (1/2)*sqrt(2)   (1/2)*sqrt(2)
 (1/2)*sqrt(2)  (-1/2)*sqrt(2)

julia> Basic.(mat(ComplexF64, H))
2×2 Matrix{Basic}:
 0.707106781186547 + 0.0*im   0.707106781186547 + 0.0*im
 0.707106781186547 + 0.0*im  -0.707106781186547 + 0.0*im
```

We want a symbolic `sqrt(2)` here, but `H` in YaoBlocks is defined with floating point numbers
There is no clean way to convert it to what we have now without manually handling it unless it is symbolic at the begining.
If we want to change, we have to think about it at a very low level, and change the design of constant gates.

GiggleLiu (approved):

lib/YaoSym/src/symengine/register.jl
```diff
@@ -23,7 +23,7 @@ function SymReg(r::AbstractArrayReg{D,<:Number}) where {D}
 end
 
 _pretty_basic(x) = x
```
GiggleLiu, line 25:
```suggestion
_pretty_basic(x) = Basic(x)
```

GiggleLiu:
The fix looks good! Thank you @wildart .

PR #356, opened 2022-04-13
most populated bit strings in an array register

GiggleLiu:
```julia
julia> reg = arrayreg(ComplexF64[0.0, 0.6, 0.8, 0.0])
ArrayReg{2, ComplexF64, Array...}
    active qudits: 2/2
    nlevel: 2

julia> most_populated(reg, 2)
2-element Vector{BitStr64{2}}:
 10 ₍₂₎
 01 ₍₂₎
```


Roger-luo:
populated seems a wrong word for this? It should be most possible?

PR #363, opened 2022-04-21
fix rotation gate dispatch

GiggleLiu:


Roger-luo:
add a test? I think this wasn't detected by current testset

GiggleLiu:
Very hard to test. They finally goes to the same instruct, no effect unless we have a specialized implementation for Rotation gates. Will become noticeable when e.g. we have a hardware required dispatching to rotation gates.

PR #367, opened 2022-04-25
better eigen basis

GiggleLiu:
Support finding the eigenbasis for terms in Rydberg Hamiltonian.


lib/YaoBlocks/src/measure_ops.jl
```diff
@@ -17,6 +21,28 @@ function eigenbasis(op::AbstractBlock{D}) where {D}
     end
 end
 
+function is_simple_diagonal(op::AbstractBlock)
+    if nqudits(op) <= 5
+        m = mat(op)
```
Roger-luo, line 26: let's just return false if the expression doesn't have any information about its matrix

Roger-luo (changes requested):

Roger-luo (approved):

PR #376, opened 2022-04-29
svg plot

GiggleLiu:


Roger-luo (approved):
I think we need to refine the doc build CLI from Bloqade but let's use this now

PR #384, opened 2022-05-05
clean up entry table by default!

GiggleLiu:
Just noticed that if I do not clean up by default, the chain getindex blow up toooooo fast!
Do you tag a new version first or you want to finish density matrix PR first? @Roger-luo


Roger-luo:
let's tag new version after density matrix

Roger-luo:
can you revert the version updates? It's easier for me to just update it with Ion before registering

GiggleLiu:
> can you revert the version updates? It's easier for me to just update it with Ion before registering

Haha, fix your ion please, it should be clever enough to detect such version issues.

Roger-luo (approved):
nvm, I think it's better to have the version bump since it's easier to declare incompatibility locally. you can merge it if it's done

PR #388, opened 2022-05-06
fix generic type

GiggleLiu:
1. Add generic typed interface to EasyBuild circuits.
2. Fix type change in getindex.

### DIFF
--- a/test/easybuild/variational_circuit.jl
+++ b/test/easybuild/variational_circuit.jl
@@ -18,16 +18,16 @@ using Yao.EasyBuild: pair_ring, pair_square, merged_rotor, rotor, rotorset
 end
 
 @testset "rotter, collect_blocks, num_gradient, opgrad" begin
-    @test  merged_rotor(true, true) == Rx(0.0)
-    @test  merged_rotor(false, false) == merged_rotor() == chain(Rz(0.0), Rx(0.0), Rz(0.0))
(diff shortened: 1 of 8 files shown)


Roger-luo (approved):

Roger-luo:
need to fix tests, rest looks good

test/easybuild/variational_circuit.jl
```diff
@@ -18,16 +18,16 @@ using Yao.EasyBuild: pair_ring, pair_square, merged_rotor, rotor, rotorset
 end
 
 @testset "rotter, collect_blocks, num_gradient, opgrad" begin
-    @test  merged_rotor(true, true) == Rx(0.0)
-    @test  merged_rotor(false, false) == merged_rotor() == chain(Rz(0.0), Rx(0.0), Rz(0.0))
-    @test  merged_rotor(false, true) == chain(Rz(0.0), Rx(0.0))
-    @test  merged_rotor(true, false) == chain(Rx(0.0), Rz(0.0))
-    @test collect_blocks(RotationGate, rotorset(:Merged, 5, true, false)) |> length == 10
+    @test  merged_rotor(Float64, true, true) == Rx(0.0)
```
Roger-luo, line 21: better set a default type rather than changing the test?

GiggleLiu: Since it is not exported, I do not think it matter I think. I like keep less interface for non-exported functions.

PR #392, opened 2022-05-19
fix density fidelity

Roger-luo:


GiggleLiu:
Hi, I feel there are some other issues related to fidelity needs to be fixed.
I find the fidelity obtained use the sqrt function has very low precision, this one might work better.

PR #393, opened 2022-05-19
fix density matrix fidelity

GiggleLiu:
* Fix density matrix fidelity
* Fix and stablize focused state fidelity.

### DIFF
--- a/lib/YaoArrayRegister/test/density_matrix.jl
+++ b/lib/YaoArrayRegister/test/density_matrix.jl
@@ -42,6 +42,7 @@ end
     reg_ = rand_state(3)
     reg2 = clone(reg, 3)
     dm = density_matrix(reg)
+    @test copy(dm) == dm
     dm_ = density_matrix(reg_)
     dm2s = density_matrix.(reg2)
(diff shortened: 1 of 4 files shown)


lib/YaoArrayRegister/test/density_matrix.jl
```diff
@@ -103,3 +104,27 @@ end
...
+    @test isapprox(expected, fidelity(r1, r2); atol=1e-5)
+
+    # focused state is viewed as mixed state
+    f1 = focus!(copy(reg1), (2, 1))
+    f2 = focus!(copy(reg2), (2, 1))
+    @test isapprox(fidelity(r1, r2), fidelity(f1, f2); atol=1e-6)
+
+    # fidelity between focused and pure state
+    f1 = rand_state(2)
+    @test isapprox(fidelity(density_matrix(f1, (1,2)), r2), fidelity(f1, f2); atol=1e-6)
```
Roger-luo, line 129: this seems worse than `sqrt`? it only has 1e-6?

Roger-luo (approved):

PR #394, opened 2022-05-29
Density matrix implementation

GiggleLiu:
* [x] partial_tr
* [x] expect
* [x] probs
* [x] measure
* [x] instruct!
* [x] UnitaryChannel
* [x] zero_state_like
* [x] algebra +, -, *, /
* [x] inplace algebra regscale!, regadd!, regsub!


GiggleLiu:
@Roger-luo I removed the method for applying a UnitaryChannel on ArrayReg to avoid confusing. Otherwise register and density matrix can have different outputs.

lib/YaoBlocks/src/composite/unitary_channel.jl
```diff
@@ -49,35 +26,39 @@ unitary_channel
...
-UnitaryChannel(it, weights) = UnitaryChannel(collect(it), weights)
-UnitaryChannel(it) = UnitaryChannel(collect(it))
+function UnitaryChannel(it, weights)
+    length(it) == 0 && error("The input operator list size can not be 0!")
+    D = nlevel(first(it))
+    UnitaryChannel(collect(AbstractBlock{D}, it), weights)
+end
 nqudits(uc::UnitaryChannel) = uc.n
 
-function YaoAPI.unsafe_apply!(r::AbstractRegister, x::UnitaryChannel)
```
Roger-luo, line 79: I think this is still needed, it's useful when you just want to sample the noise

GiggleLiu: It creates two inconsistent behaviors.
1. The register simulation and density matrix simulation can give different expectation values.
2. Since we have removed its mat implementation, we do not have composibility of gates. e.g. put(5, 2=>channel) does not work.

It would suggest adding a new block: Probabilistic in the future rather than the UnitaryChannel, then it should be fine to get its matrix representation. Channel is more likely for the density matrix.

PR #396, opened 2022-06-01
Fix google supremacy circuit

GiggleLiu:


lib/YaoBlocks/src/composite/unitary_channel.jl
```diff
@@ -1,63 +1,49 @@
...
 """
 struct UnitaryChannel{D, W<:AbstractVector} <: CompositeBlock{D}
     n::Int
     operators::Vector{AbstractBlock{D}}
-    weights::W
+    probs::W
 
     function UnitaryChannel(operators::Vector{AbstractBlock{D}}, w::AbstractVector) where D
         @assert length(operators) == length(w) && length(w) != 0
         if !(sum(w) ≈ 1)
```
Roger-luo, line 20: should we also check they are positive? since it's probabilities now

Roger-luo (approved):
rest LGTM. the check is not blocking, but would be nice to have

lib/YaoBlocks/src/composite/unitary_channel.jl
GiggleLiu: 👍🏾

PR #398, opened 2022-06-03
[unitaryhack] Example on Riemannian gradient-flow optimizer

radras:


GiggleLiu:
Nice work! Currently this file does not build successfully:
```julia
[ Info: generating markdown page from `~/.julia/dev/Yao/docs/src/examples/8.riemannian-gradient-flow/main.jl`
ERROR: LoadError: LoadError: ArgumentError: Package KrylovKit not found in current path.
- Run `import Pkg; Pkg.add("KrylovKit")` to install the KrylovKit package.
in expression starting at /home/leo/.julia/dev/Yao/docs/src/generated/examples/8.riemannian-gradient-flow/index.md:2
when executing the following code block from inputfile `~/.julia/dev/Yao/docs/src/examples/8.riemannian-gradient-flow/main.jl`

using Yao, Yao.EasyBuild, Plots, Random
using KrylovKit: eigsolve
```

You probably want to add KrylovKit to the dependency. 

@Roger-luo I feel as the numer of examples grow, the build is really slow. Do you have recommended development workflow?

Roger-luo:
> @Roger-luo I feel as the numer of examples grow, the build is really slow. Do you have recommended development workflow?

Let's wait for the CI/CD work on Bloqade/QuEra first, then we can use the hardware and CI from there.

GiggleLiu:
@radras , currently, you can debug your notebook with the following command:

```bash
$ bash docs/servedoc.sh
```

This liveserver will update the web page automatically once you changed an example.

Roger-luo:
I think this is in general good to go, very well written, I'd like to merge this first. Thanks! (I've put the PR as resolved a few weeks ago for you to receive the bonuty, please let me know if you didn't receive it)

radras:
Sorry, I made a mistake in the last commit, I'll fix it now.

GiggleLiu:
I also have a question during reviewing this PR, I can only get the gradient of $U$ as $2\rho U^\dagger H$. Can you please also write more explicitly how you get this gradient in the new commit? I am very curious about it.

radras:
How did you derive it? They have the derivation in Appendix C here https://arxiv.org/pdf/2202.06976.pdf

GiggleLiu:
Now I see the problem. You'd better not using notation of $\text{grad}\mathcal{L}(U)$ to avoid potential confusing.
It is different from (unconstraint) matrix gradient because in the paper, they also add a requirement that the $U$ is unitary, then the gradient must be projected to the tangent space. This is why they anti-symmetrize the gradient using the commutator.

radras:
I made a new PR #414

GiggleLiu:
> What do you suggest, should I maybe use nabla? Since this paper is the main reference I wouldn't necessarily change the notation, maybe I can just make a brief comment pointing to the derivation in the appendix or something like that?

About the notation, there are some aspects can be improved (sorry for being picky on notations).
Since you are doing gradient descent, I do not think using $\dot{U}$ kind of notation is good, it looks like time evolution.
I think you can use the word adjoint to denote the gradient: $\overline{U}$. It is fine to keep the original notation, I find using dot is quite confusing while reading throught it, but finally I understand what you wanted to present.

For the Riemanian gradient, you can add some description like: "This adjoint of U is not is different from the adjoint obtained by treating it as a regular matrix. It can be understood as the matrix adjoint projected to the tangent space of $U$."

PR #402, opened 2022-06-09
Fix Yao Blocks

GiggleLiu:
fix docstrings and a fallback


GiggleLiu:
Fix multiple documentation error and a dispatch. I need to get this PR merged to master so that people won't be confused that much. We'd better tag a patch version later. Can you check if these changes are correct later? @Roger-luo

PR #404, opened 2022-06-10
fix merging empty entry table

GiggleLiu:


GiggleLiu:
I will merge first, you can still review later.

PR #415, opened 2022-07-08
fix ad error for abstract vector constructors

GiggleLiu:
The `sum` and `put` constructors are still not properly supported.
Zygote dispatches these functions to incorrect backward rules.
I do not have time to fix all interfaces, after all, the AD engine should be able to figure out the incorrect backward rules specification.

@VarLad


Roger-luo (approved):
LGTM

PR #416, opened 2022-07-13
support basic error channels

Roger-luo:
moved `DensityMatrix` back to `YaoArrayRegister` to fix type piracy


lib/YaoAPI/src/registers.jl
```diff
@@ -707,17 +707,6 @@ end
 
 ####################### Density Matrix ############
 
-"""
-    DensityMatrix{D,T,MT<:AbstractMatrix{T}} <: AbstractRegister{D}
```
GiggleLiu, line 711: Why do you move DensityMatrix out from YaoAPI? I think it will be convenient for developers to extend if we put it here.

lib/YaoArrayRegister/src/density_matrix.jl
```diff
@@ -15,7 +28,36 @@ YaoAPI.nqubits(ρ::DensityMatrix) = nqudits(ρ)
...
+    `isapprox(dm.state, dm.state')` or explicit mark it as `Hermitian`.
+"""
+function rand_density_matrix(n::Int; nlevel::Int=2, pure::Bool=false)
+    return rand_density_matrix(ComplexF64, n; nlevel, pure)
+end
+
+function rand_density_matrix(::Type{T}, n::Int; nlevel::Int=2, pure::Bool=false) where T
+    pure && return density_matrix(rand_state(T, n; nlevel))
+
+    N = nlevel^n
```
GiggleLiu, line 51:
```julia
julia> rand_density_matrix(ComplexF64, 3)
DensityMatrix{2, Float64, Matrix{Float64}}([0.09895989446363539 0.033083525024186204 … 0.024417543183544423 0.0015611307789331337; 0.033083525024186225 0.09796262397994808 … -0.01405934647515636 -0.009407688198002211; … ; 0.024417543183544395 -0.014059346475156385 … 0.152166399588001 -0.018906510990330864; 0.0015611307789331207 -0.00940768819800221 … -0.018906510990330857 0.13113192677153102])
```

Data type is not correct. Why don't we just do
```julia
partial_tr(density_matrix(rand_state(6; nlevel=2)), (1,2,3))
```

lib/YaoBlocks/src/channel/error.jl
```diff
@@ -0,0 +1,66 @@
...
+```
+
+### Arguments
+
+- `n`: number of qubits.
+- `p`: probability of bit flips.
+- `locs`: locations of qubits that has this error to occur.
+"""
+function bit_flip_channel(n::Int, p::Real, locs::NTuple{N, Int}) where N
+    opX = repeat(n, X, locs)
```
GiggleLiu, line 17: bit flip channel is single bit flip channel right? this error channel flips all bits. I guess this is wrong?

lib/YaoBlocks/src/channel/error.jl
```diff
@@ -0,0 +1,66 @@
...
+p⋅ρ + (1-p)⋅ZρZ
+```
+
+### Arguments
+- `n`: number of qubits.
+- `p`: probability of this error to occur.
+- `locs`: locations of qubits that has this error to occur.
+"""
+function phase_flip_channel(n::Int, p::Real, locs::NTuple{N, Int}) where N
+    opZ = repeat(n, Z, locs)
```
GiggleLiu, line 36: The same here.

lib/YaoBlocks/src/channel/error.jl
```diff
@@ -0,0 +1,66 @@
...
+
+struct DepolarizingChannel{T} <: PrimitiveBlock{2}
+    n::Int # n is not necessary but this is required by a block
+    p::T
+end
+
+YaoAPI.nqudits(ch::DepolarizingChannel) = ch.n
+
+function YaoAPI.unsafe_apply!(dm::DensityMatrix, ch::DepolarizingChannel)
+    regscale!(dm, 1 - ch.p)
```
GiggleLiu, line 48: The same here, I think we need to have a location parameter to specify which qubits have such error.

GiggleLiu (changes requested):

lib/YaoAPI/src/registers.jl
Roger-luo: this is causing type piracy in `YaoArrayRegister`

lib/YaoBlocks/src/channel/error.jl
Roger-luo: The depolarizing channel doesn't have a location by definition tho. It's just adding identity to the entire density matrix.

lib/YaoAPI/src/registers.jl
GiggleLiu: Yeah, this can not be avoided right, we already choose such design pattern.
I guess we have a lot other piracy already.

lib/YaoArrayRegister/src/density_matrix.jl
GiggleLiu: Why it causes too much memory? It might be a bug in `partial_tr`

Roger-luo: I think you are right, change it back

lib/YaoAPI/src/registers.jl
GiggleLiu: > I don't see anyone is using YaoAPI to extend density matrix after all

The same reason applies here: I don't see anyone is suffering from this type of piracy. DM is just general purposed, we hope whoever what to extend Yao's density matrix can have a very light weighted API package to extend. This is the purpose of YaoAPI. We allow piracy because we consider YaoAPI as kind of Yao's submodule. The piracy here is not a problem because we do not implement any function on DM type in the API package.

lib/YaoBlocks/src/channel/error.jl
```diff
@@ -0,0 +1,115 @@
...
+            else
+                p
+            end
+        end
+    end
+end
+
+function single_qubit_error_composite_operators(err, n::Int, locs::NTuple)
+    total = length(locs)
+    return map(0:1<<total-1) do syndrome
```
GiggleLiu, line 15: I am not sure you have to create exponential many syndromes. Can we do it in a more clever way?

lib/YaoBlocks/src/channel/error.jl
GiggleLiu: If you add a `locs` argument, you can decide which qubit has the depolarizing error. Then their state will become completely mixed. No?

lib/YaoBlocks/test/channel/error.jl
```diff
@@ -0,0 +1,53 @@
+using Test
```
GiggleLiu, line 1: I guess you forgot to include this file in tests.

lib/YaoBlocks/src/channel/error.jl
Roger-luo: This is during construction and only exponential on number of locations. This has no difference with applying a list of channels. I think this is more convenient.

Roger-luo: kron of list of unitary channels is more costly than just compose them into one.

lib/YaoAPI/src/registers.jl
GiggleLiu: Interesting, I want to hear more about the method saving mechanism in Julia. Let us discuss in person.

lib/YaoBlocks/src/channel/error.jl
jlbosse: There is indeed a sensible definition of a multi-qubit depolarizing channel 
and it can be written down as a unitary channel as follows:
$$
  ρ ↦ (1-p) ρ + p/(2^k - 1) ∑_{P_k ≠ id} P_k ρ P_k
$$
where the sum runs over all non-identity Pauli-strings acting on qubits 1
to k and ρ is a n-qubit density matrix. For simplicity, I assumed that we want 
the depolarizing channel on the first k qubits, but it should be clear how 
this definition extends to the general case where we want depolarizing noise on 
an arbtitrary subset of k qubits.

However, this definition leads to 2^k Kraus operators in the unitary channel 
and is hence fairly slow to implement (unless k=n). For this reason QuEST 
only provides the one and two-qubit depolarizing channels. See [here](https://quest-kit.github.io/QuEST/group__decoherence.html)

For this reason, I would suggest keeping the global depolarizing channel as it 
is currently defined and add a function 
```julia
function single_qubit_depolarizing_channel(n, loc, p)
    return pauli_error_channel(n, loc, px=p/3, py=p/3, pz=p/3)
end
```
and a similarly defined `two_qubit_depolarizing_channel(n, locs, p)`.

Roger-luo: @GiggleLiu Does this sound good?

GiggleLiu: Yes, it makes sense. 👍🏾

Roger-luo: Ok now if we agree on this, probably just go ahead fork my branch to put your proposal in. I can review it. Thanks! @jlbosse

jlbosse: Will do! And is the debate about type piracy and where to put DensityMatrix resolved in the current version?

Roger-luo: I think given now in 1.9 native code caching, we need and should eliminate type piracy and more specifically invalidations as much as possible to utilize native code caching. But we can do that in a separate PR. Just go ahead to include this first.

PR #417, opened 2022-07-19
Clarified documentation of @const_gate by adding a description of the nlevel=D option

jlbosse:
I had to read the sources to find out how to create a const_gate that acts on qutrits instead of qubits. To make it simpler for the next person I added some details to the docstring  of the const_gate macro.


GiggleLiu (commented):
Good catch, thank you!

GiggleLiu (approved):

PR #420, opened 2022-08-15
safely remove qudits from a register

GiggleLiu:
* `isseparable` returns true if a quantum state is separable.
* `join` can be join two density matrices now,
* `safe_remove!` function can be used to remove ancilla qudits that separable from other qudits,


lib/YaoArrayRegister/src/operations.jl
```diff
@@ -211,3 +211,39 @@ end
...
+    else
+        return rank(reshape(r.state, D^p, :)) == 1
+    end
+end
+
+"""
+$(TYPEDSIGNATURES)
+
+Remove qubits that are not entangled with the rest qudits safely.
+i.e. `isseparable(reg, locs)` must return true.
```
Roger-luo, line 240: Maybe also have an example here?

Roger-luo (approved):

PR #421, opened 2022-08-15
new reflection gate

GiggleLiu:
In this PR, I define the `Projector` as $$|s\rangle \rightarrow |s\rangle \langle s|$$,
`ReflectGate` as the time evolution over `Projector`: $$|v⟩ → 1 - (1-\exp(-iθ)) |v⟩⟨v|$$, where θ is the evolution time.


Roger-luo (commented):
Only one question what's the difference between the old reflect gate and this one?

GiggleLiu:
This one is an alias of the time evolution over the reflection gate, while the old one is implemented as a new type.
This new one is also more general, the evolution time $\theta$ can be not $\pi$.

Roger-luo (approved):

PR #427, opened 2022-09-30
Removed DocStringExtensions from dependencies of YaoBlocks

jlbosse:
YaoBlocks was blocking the update of DocStringExtensions to v0.9. After closer inspection, DocStringExtensions was used very sparingly, so I manually inserted all the corresponding docstrings and removed the dependency


jlbosse:
Sorry, somehow older commits still showed up here. Will post an updated PR in a sec

GiggleLiu:
Hi, can you please fix the dependency instead removing it. We want to switch to DocStringExtensions gradually for better consistency between function signatures and docstrings.

PR #435, opened 2022-10-28
fix issue 432

### LINKED ISSUE #432: `expect(::AbstractBlock, ::ArrayReg)` is not recognized as typestable
Currently the output type of `expect(::PutBlock, ::ArrayReg)` is inferred to be `Union{ArrayReg{...}, Complex}` and the output
type of `expect(::AddBlock, ::ArrayReg)` is inferred to be `Any`. This is a problem in one of my applications, because I wanted to use `return_types(expect, Tuple{typeof(hamiltonian),typeof{register}})` to preallocate some memory. See the following MWE to confirm the non-inferability:

```julia
using Yao

nq = 2
O1 = put(nq, 1 => X)
O2 = 0.5 * put(nq, 2 => Y)
H = O1 + O2
reg = rand_state(nq)

@code_warntype expect(O1, reg)

# output
MethodInstance for YaoAPI.expect(::PutBlock{2, 1, XGate}, ::ArrayReg{2, ComplexF64, Matrix{ComplexF64}})
  from expect(op::AbstractBlock, reg::AbstractRegister) in YaoBlocks at .../Yao/lib/YaoBlocks/src/blocktools.jl:112
Arguments
  #self#::Core.Const(YaoAPI.expect)
  op::PutBlock{2, 1, XGate}
  reg::ArrayReg{2, ComplexF64, Matrix{ComplexF64}}
Body::Union{ArrayReg{2, ComplexF64, Matrix{ComplexF64}}, ComplexF64}
1 ─ %1 = YaoBlocks.:var"'"(reg)::AdjointRegister{2, ArrayReg{2, ComplexF64, Matrix{ComplexF64}}}
│   %2 = YaoBlocks.copy(reg)::ArrayReg{2, ComplexF64, Matrix{ComplexF64}}
│   %3 = YaoBlocks.apply!(%2, op)::ArrayReg{2, ComplexF64, Matrix{ComplexF64}}
│   %4 = (%1 * %3)::Union{ArrayReg{2, ComplexF64, Matrix{ComplexF64}}, ComplexF64}
└──      return %4

@code_warntype expect(H, reg)

# output

MethodInstance for YaoAPI.expect(::Add{2}, ::ArrayReg{2, ComplexF64, Matrix{ComplexF64}})
  from expect(op::AbstractAdd, reg::AbstractRegister) in YaoBlocks at .../Yao/lib/YaoBlocks/src/blocktools.jl:165
Arguments
  #self#::Core.Const(YaoAPI.expect)
  op::Add{2}
  reg::ArrayReg{2, ComplexF64, Matrix{ComplexF64}}
Locals
  #194::YaoBlocks.var"#194#195"{ArrayReg{2, ComplexF64, Matrix{ComplexF64}}}
Body::Any
1 ─ %1 = YaoBlocks.:(var"#194#195")::Core.Const(YaoBlocks.var"#194#195")
│   %2 = Core.typeof(reg)::Core.Const(ArrayReg{2, ComplexF64, Matrix{ComplexF64}})
│   %3 = Core.apply_type(%1, %2)::Core.Const(YaoBlocks.var"#194#195"{ArrayReg{2, ComplexF64, Matrix{ComplexF64}}})
│        (#194 = %new(%3, reg))
│   %5 = #194::YaoBlocks.var"#194#195"{ArrayReg{2, ComplexF64, Matrix{ComplexF64}}}
│   %6 = YaoBlocks.sum(%5, op)::Any
└──      return %6
```

After some digging the culprit seems to be that `*(::AdjointRegister{D,<:ArrayReg}, ::ArrayReg{D})` 
is not type stable. This is the code in question: 
https://github.com/QuantumBFS/Yao.jl/blob/4bc2a6ff18970ded584190732c722a8a997cbee5/lib/YaoArrayRegister/src/operations.jl#L161-L171

IMHO `*(bra::AdjointRegister, ket::ArrayReg)` should always return an `ArrayReg`, albeit one on zero qudits if the dimensions match. This is similar to the case in LinearAlgebra:

```julia
using LinearAlgebra

v = rand(10, 1)
v' * v

# output is a matrix, not a float.
1×1 Matrix{Float64}
 2.584654 
```

and there should be another method `dot(::ArrayReg, ::ArrayReg)` that is guaranteed to return a number and not a register.

However, I am not deep enough into the code base, to understand how feasible that approach would be (and in particular if there are cases where `*(::AdjointReg, ::ArrayReg)` gets called and it is not clear 
whether a number or a register is expected as the output. And this change would be a breaking change, if any users rely on the current behaviour.

Another, simpler fix would be to annotate ``expect(op::AbstractBlock{D}, reg::AbstractArrayReg{D,T,AT})::T where {D,T,AT}`` to fix the output type of expect to be the data type of the register. Again, I am not deep enough in the code, to understand whether there are cases where this currently is not the case or where this behaviour would be undesirable.

GiggleLiu:
fix #432 

## Change List
* New exported API `sandwich` in YaoBlocks for computing `<bra| op |ket>`.
* `expect` function now returns real and errors on complex values.
* The multiplication `<bra| * |ket>` now returns a number, and throws an error if the qubit number does not match.

## Fixes the type stability issue
```julia
julia> using Yao
[ Info: Precompiling Yao [5872b779-8223-5990-8dd0-5abbb0748c8c]

julia> nq = 2
2

julia> O1 = put(nq, 1 => X)
nqubits: 2
put on (1)
└─ X

julia> reg = rand_state(nq)
ArrayReg{2, ComplexF64, Array...}
    active qubits: 2/2
    nlevel: 2

julia> @code_warntype expect(O1, reg)
MethodInstance for YaoAPI.expect(::PutBlock{2, 1, XGate}, ::ArrayReg{2, ComplexF64, Matrix{ComplexF64}})
  from expect(op::AbstractBlock, reg::AbstractRegister) in YaoBlocks at /home/leo/.julia/dev/Yao/lib/YaoBlocks/src/blocktools.jl:95
Arguments
  #self#::Core.Const(YaoAPI.expect)
  op::PutBlock{2, 1, XGate}
  reg::ArrayReg{2, ComplexF64, Matrix{ComplexF64}}
Body::Float64
1 ─ %1 = YaoBlocks.sandwich(reg, op, reg)::ComplexF64
│   %2 = Base.broadcasted(YaoBlocks.safe_real, %1)::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0}, Nothing, typeof(YaoBlocks.safe_real), Tuple{ComplexF64}}
│   %3 = Base.materialize(%2)::Float64
└──      return %3
```

## Examples
```julia
julia> expect(O1, reg)
0.05754647503710675

julia> sandwich(reg, O1, reg)
0.05754647503710675 + 0.0im
```
## Note

For `expect(O1 + O2, reg)`, the code is still type unstable, this is due to the `Add` block contains a vector of abstract type.


lib/YaoArrayRegister/src/operations.jl
```diff
@@ -158,40 +158,46 @@ for op in [:*, :/]
...
+
+julia> reg2 = uniform_state(3);
+
+julia> reg1' * reg2
+0.5 + 0.0im
+```
+"""
+function Base.:*(bra::AdjointRegister{D,<:ArrayReg}, ket::ArrayReg{D})::Number where D
+    # check the register sizes
+    nqudits(bra) == nqudits(ket) && nremain(bra) == nremain(ket) || error(
```
Roger-luo, line 179: do you remember what was the use case for contracting active qubits only? it seems losing functionality if we let this error

lib/YaoArrayRegister/src/operations.jl
```diff
@@ -158,40 +158,46 @@ for op in [:*, :/]
...
+            "partially contract ⟨bra|ket⟩ is not supported, expect ⟨bra| and |ket⟩ to have the same size. Got nactive(bra)/nqudits(bra)/nbatch(bra)=$(nactive(bra))/$(nqudits(bra))/$(nbatch(bra)), nactive(ket)/nqudits(ket)=$(nactive(ket))/$(nqudits(ket))/$(nbatch(ket))",
         )
+    A, C = parent(state(parent(bra))), parent(state(ket))
+    res = zeros(eltype(promote_type(T1, T2)), nbatch(ket))
+    for j = 1:size(A, 2)
+        for i = 1:size(A, 1)
+            @inbounds res[i] += conj(A[i, j]) * C[i, j]
+        end
     end
+    res
```
Roger-luo, line 200:
```suggestion
    res = zeros(eltype(promote_type(T1, T2)), nbatch(ket))
    for j in 1:size(A, 2), i in 1:size(A, 1)
        @inbounds res[i] += conj(A[i, j]) * C[i, j]
    end
    return res
```

lib/YaoBlocks/src/utils.jl
```diff
@@ -311,3 +311,12 @@ end
 
 SparseArrays.sparse(et::EntryTable) = SparseVector(et)
 Base.vec(et::EntryTable) = Vector(et)
+
+# convert a (maybe complex) number x to real number.
+function safe_real(x)
+    img = imag(x)
+    if !(iszero(img) || isapprox(x - im*img, x))
```
Roger-luo, line 318: why not just `isapprox(img, zero(img))`? It covers both cases (exact zero or approx zero)

also this line is not tested.

lib/YaoBlocks/src/utils.jl
```diff
@@ -311,3 +311,12 @@ end
...
+        error("Can not convert number $x to real due to its large imaginary part.")
```
Roger-luo, line 319: this should be an `InexactError` and the name of this function should be `approx_real`? `safe_real` sounds more like those `unsafe_xxx` families in `Base` or for integers.

Roger-luo (changes requested):
LGTM in general, only minor changes is needed.

lib/YaoArrayRegister/src/operations.jl
Roger-luo: or there should be a `contract` function that always return a `Matrix`.

lib/YaoBlocks/src/utils.jl
GiggleLiu: This is a bit tricky. I want to support float, complex and Basic types.
`isapprox(img, zero(img))` is also too strict because there can be small imaginary part for real numbers.

GiggleLiu: Wield, I can not see which line is not covered from the files tab.

lib/YaoBlocks/src/utils.jl
GiggleLiu: Sounds good

lib/YaoArrayRegister/src/operations.jl
GiggleLiu: It is not intuitive to return a matrix, because it is hard to explain to users.
If we need it as a developer, we can use `state(bra) * state(ket)` directly.
Here, I prefer to make it simple. Either error on not fully activated registers, or not distinguishing on active and remaining qubits.

Roger-luo: Yeah I think just using `state` makes sense.

lib/YaoBlocks/src/utils.jl
Roger-luo: Do you have an example case that `isapprox(x, zero(x))` fails? It's not very intuitive to me that there's a difference between these two implementations.

GiggleLiu:
I think this PR should be merged. Automatic differentiating complex numbers does not make sense.

lib/YaoBlocks/src/utils.jl
GiggleLiu: `eps(0.0)` is too small.

```julia
julia> isapprox(imag(1+1e-17im), 0.0)
false

julia> eps(0.0)
5.0e-324
```

Roger-luo (approved):

PR #446, opened 2022-12-07
Changed expect_g and fidelity_g to return the "correctly" scaled gradient w.r.t to the register

jlbosse:
Up until now `expect_g` and `fidelity_g` were off by a factor of one half, in the sense that 
∂_reg expect(ham, reg) = 2 * expect_g(ham, reg) instead ∂_reg expect(ham, reg) = expect_g(ham, reg) and analogous for the other methods of `expect_g` and `fidelity_g`. This was counter acted by multiplying a factor of 2 to the final `paramsδ` at the end of `expect_g(ham, reg => circuit)` and `fidelity_g(reg1 => circuit1, reg2 => circuit2)`.

This is changed here to backpropagate the "correctly" scaled gradient w.r.t registers.

I am not sure if this would be considered a breaking change, because it also means that the output of `expect'(ham, reg => circuit)[1]` is changed by a factor of 2 now. 

Changes: 
 - Changed expect_g and fidelity_g
 - Updated the rrules in chainrules_patch.jl
 - Updated the corresponding tests

### DIFF
--- a/lib/YaoBlocks/src/autodiff/chainrules_patch.jl
+++ b/lib/YaoBlocks/src/autodiff/chainrules_patch.jl
@@ -122,7 +122,7 @@ function rrule(
     function (outδ)
         reg, c = reg_and_circuit
         out = copy(reg) |> c
-        goutreg = copy(out) |> op
+        goutreg = 2copy(out) |> op
         for b = 1:YaoArrayRegister._asint(nbatch(goutreg))
             viewbatch(goutreg, b).state .*= outδ[b]
         end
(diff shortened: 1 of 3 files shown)


Roger-luo:
need tests pass before a review

GiggleLiu (approved):
Looks good to me. The failure in the nightly test seems unrelated.

jlbosse:
The tests fun on my local machines on julia 1.8.3 and 1.8.2 and are in parts of the tests that shouldn't be affected by my changes at all. So I also think that the failures on nightly are unrelated.

PR #452, opened 2023-01-12
Added similar(::DensityMatrix)

jlbosse:
The title says it all. I also checked that #416 and #408 don't add `similar(::DensityMatrix)`


Roger-luo:
LGTM

Roger-luo (approved):

Roger-luo (commented):
Could you give a short description on what's the use case of this?

jlbosse:
I have code where I need a "scratch" register to store a temporary state. I was allocating creating this scratch register using `scratch = similar(register)` and when I wanted to run a density matrix simulation (with noise channels added to the circuit) instead of a noiseless state vector simulation it was complaining about the lack of `similar(::DensityMatrix)`

PR #454, opened 2023-01-29
add `circuit_depth`

### LINKED ISSUE #426: Implementation of function to calculate circuit depth
Request for a function equivalent to `circuit.depth()` in [Qiskit](https://qiskit-test.readthedocs.io/en/latest/api/qiskit.circuit.QuantumCircuit.html?highlight=depth#qiskit.circuit.QuantumCircuit.depth).

[Here](https://quantumcomputing.stackexchange.com/a/5772) is the logic to find the depth, note that since at each step, the depth of the `AbstractBlock` needs to be added, which may not just be 1.

I may have missed this in the docs and the issues, or it may be called differently, and if so, please close the issue.

DhruvaSambrani:
Fixes #426


Roger-luo (commented):
can we have a test?

Roger-luo:
Yes I think it should be a keyword argument since it's an option, thanks!

Roger-luo:
need to fix the test and make sure it passes the CI, please let me know if you have issues with that

DhruvaSambrani:
yes sure

GiggleLiu:
Thanks for the pull request, but I think the depth counting should be implemented in a recursive style.
Suppose I have a nested chain, the counting of depth might be wrong.

You should add a test with the following circuit:

```julia
chain(chain(put(4, 2=>X), control(4, 2, 3=>Z)))
```

PR #455, opened 2023-02-16
More work on basic error channels

jlbosse:
Continuation of #416 with added documentation, tests and single and two-qubit depolarizing channels as well as getting up to date with master.


jlbosse:
Before this gets merged, I was wondering if it might make more sense to define the different error channels without wrapping the unitaries in `PutBlock`s. This allows to wrap the output of `bit_flip_channel` in a `PutBlock` or `KronBlock` and have the same error channel act on many different locations in the circuit. So do 
```julia
function bit_flip_channel(p)
    return unitary_channel([igate(1), X], [1-p, p])
end
```
instead of
```julia
function bit_flip_channel(n::Int, loc::Int; p::Real)
    return UnitaryChannel([igate(n), put(n, loc=>X)], [p, 1-p])
end
```

jlbosse:
The CI failure seems unrelated to me and locally compiled docs also look good. So I think this is ready for a review @Roger-luo or @GiggleLiu !

lib/YaoArrayRegister/src/density_matrix.jl
```diff
@@ -1,5 +1,18 @@
...
+    DensityMatrix{D,T,MT<:AbstractMatrix{T}} <: AbstractRegister{D}
+    DensityMatrix{D}(state::AbstractMatrix)
+    DensityMatrix(state::AbstractMatrix; nlevel=2)
+
+Density matrix type, where `state` is a matrix.
+Type parameter `D` is the number of levels, it can also be specified by a keyword argument `nlevel`.
+"""
+struct DensityMatrix{D,T,MT<:AbstractMatrix{T}} <: AbstractRegister{D}
+    state::MT
+end
```
GiggleLiu, line 12: What is the purpose of moving `DensityMatrix` from `YaoAPI` to `YaoArrayRegister`? The reason why we let the elementary types and basic functions be defined in `YaoAPI` is for the easy extension purpose. Other packages only need to depend on YaoAPI.

lib/YaoArrayRegister/src/density_matrix.jl
```diff
@@ -16,7 +29,50 @@ YaoAPI.nqubits(ρ::DensityMatrix) = nqudits(ρ)
...
+
+    The generated density matrix is not strict hermitian due to rounding error.
+    If you need to check hermicity, do not use `ishermitian` consider using
+    `isapprox(dm.state, dm.state')` or explicit mark it as `Hermitian`.
+"""
+function rand_density_matrix(n::Int; nlevel::Int=2, pure::Bool=false)
+    return rand_density_matrix(ComplexF64, n; nlevel, pure)
+end
+
+function rand_density_matrix(::Type{T}, n::Int; nlevel::Int=2, pure::Bool=false) where T
```
GiggleLiu, line 49: It is not recommended to initialize random density matrix with partial trace. It will allocate you a huge memory.
```julia
julia> dm = rand_density_matrix(10)
ERROR: OutOfMemoryError()
Stacktrace:
  [1] Array
    @ ./boot.jl:461 [inlined]
  [2] Array
    @ ./boot.jl:469 [inlined]
  [3] similar
    @ ./array.jl:378 [inlined]
  [4] similar
    @ ~/.julia/juliaup/julia-1.8.5+0.x64.linux.gnu/share/julia/stdlib/v1.8/LinearAlgebra/src/adjtrans.jl:212 [inlined]
  [5] *
    @ ~/.julia/juliaup/julia-1.8.5+0.x64.linux.gnu/share/julia/stdlib/v1.8/LinearAlgebra/src/matmul.jl:154 [inlined]
  [6] density_matrix(reg::ArrayReg{2, ComplexF64, Matrix{ComplexF64}})
    @ YaoArrayRegister ~/.julia/dev/Yao/lib/YaoArrayRegister/src/density_matrix.jl:81
  [7] rand_density_matrix(::Type{ComplexF64}, n::Int64; nlevel::Int64, pure::Bool)
    @ YaoArrayRegister ~/.julia/dev/Yao/lib/YaoArrayRegister/src/density_matrix.jl:51
  [8] #rand_density_matrix#60
    @ ~/.julia/dev/Yao/lib/YaoArrayRegister/src/density_matrix.jl:46 [inlined]
  [9] rand_density_matrix(n::Int64)
    @ YaoArrayRegister ~/.julia/dev/Yao/lib/YaoArrayRegister/src/density_matrix.jl:45
 [10] top-level scope
    @ REPL[21]:1
```

lib/YaoArrayRegister/src/density_matrix.jl
```diff
@@ -16,7 +29,50 @@ YaoAPI.nqubits(ρ::DensityMatrix) = nqudits(ρ)
...
+    indent = get(io, :indent, 0)
+    summary(io, dm)
+    println(io)
+    println(io, " "^(indent+2), "state:")
+    s = sprint(dm.state) do io, x
+        show(io, mime, x)
+    end
+    s = map(s->" "^(indent+4) * s, split(s, '\n'))
+    join(io, s[2:end], '\n')
+end
```
GiggleLiu, line 71: The line `state:` looks wield, but this is a minor issue.
I think we can just reuse the `print_matrix` function in Julia Base. It will print large matrix properly with dots.
```julia
julia> Base.print_matrix(IOContext(stdout, :limit => true), round.(randn(ComplexF64, 400, 400); digits=5))
 -0.76732 - 0.59468im  -0.10658 - 0.95842im  …   0.34798 + 0.7456im
  0.86556 - 1.11837im  -0.27888 + 0.0089im      -0.20812 + 0.09023im
 -0.83708 - 0.02078im  -0.94368 - 0.87756im     -1.86103 + 0.11575im
 -0.92029 + 0.19052im   1.82818 - 0.7781im        0.0909 + 0.79133im
  -0.0152 + 0.04503im  -0.59895 + 0.2377im       1.02229 + 0.11843im
  0.73325 + 0.39854im   0.55827 - 1.26088im  …   0.07015 + 0.07076im
  0.14204 - 0.13815im   0.14383 + 0.12192im      0.05521 + 0.32841im
  0.13662 + 0.81992im   0.39468 + 0.28733im     -0.44734 + 0.91129im
 -1.19599 - 0.13676im   0.69781 + 0.8893im       0.72188 - 0.02428im
 -0.45704 - 1.50642im   0.44714 - 0.34654im      0.32411 + 0.05651im
          ⋮                                  ⋱  
  0.16856 - 0.67063im   0.06789 - 0.72864im     -0.84355 + 0.12563im
 -0.13521 + 0.90965im  -0.30472 + 0.30413im     -0.29758 - 0.56964im
  -0.6075 + 0.93425im   0.02701 + 0.14438im      1.84863 - 0.21101im
  0.63625 + 0.4728im   -0.52869 + 0.12177im     -0.05003 - 0.90055im
  0.23096 + 0.48462im   0.33513 - 0.24298im  …   0.92307 - 1.14345im
    0.099 + 0.73452im   0.90888 + 0.09793im     -0.15182 + 0.57916im
  0.28379 - 0.35861im   0.07362 + 1.31831im      1.31751 + 0.67337im
  0.18581 + 2.01336im     0.382 - 0.0502im      -0.63735 + 0.1921im
  0.59116 + 0.21242im  -0.73689 - 0.28261im      0.67455 - 0.06213im
```

lib/YaoArrayRegister/src/density_matrix.jl
```diff
@@ -16,7 +29,50 @@ YaoAPI.nqubits(ρ::DensityMatrix) = nqudits(ρ)
...
+
+# Density matrices are hermitian
+Base.adjoint(dm::DensityMatrix) = dm
+LinearAlgebra.ishermitian(dm::DensityMatrix) =  true
```
GiggleLiu, line 75: `ishermitian` is for operators. Do we need this API for now? I can not come up with a using case.

lib/YaoArrayRegister/test/density_matrix.jl
```diff
@@ -135,6 +135,18 @@ end
     # fidelity between focused and pure state
     f1 = rand_state(2)
     @test isapprox(fidelity(density_matrix(f1, (1,2)), r2), fidelity(f1, f2); atol=1e-6)
+
+    dm = rand_density_matrix(2)
```
GiggleLiu, line 139: Some lines in the source code are not covered by this test.

lib/YaoBlocks/src/channel/error_channel.jl
```diff
@@ -0,0 +1,102 @@
...
+as `kron(single_qubit_depolarizing_channel(p), single_qubit_depolarizing_channel(p))`.
+"""
+function two_qubit_depolarizing_channel(p::Real)
+    return UnitaryChannel(
+        [kron(I2, I2), kron(I2, X), kron(I2, Y), kron(I2, Z),
+         kron(X, I2), kron(X, X), kron(X, Y), kron(X, Z),
+         kron(Y, I2), kron(Y, X), kron(Y, Y), kron(Y, Z),
+         kron(Z, I2), kron(Z, X), kron(Z, Y), kron(Z, Z),
+        ],
+        [1-15p/16, p/16, p/16, p/16,
```
GiggleLiu, line 96: I am a bit worried about the performance. The definition looks very simple. Can we have a better implementation? We can keep an issue if there is no quick solution to it.
```math
(1-\lambda )\rho +{\frac {\lambda }{d}}I
```

lib/YaoBlocks/test/channel/error_channel.jl
```diff
@@ -0,0 +1,46 @@
...
+@testset "pauli channel" begin
+    dm = rand_density_matrix(3)
+    ch = put(3, 1 => pauli_error_channel(pz=0.2, py=0.1, px=0.3))
+    opX = mat(put(3, 1=>X))
+    opY = mat(put(3, 1=>Y))
+    opZ = mat(put(3, 1=>Z))
+
+    @test 0.4 * dm.state + 0.3 * opX * dm.state * opX + 
+        0.1 * opY * dm.state * opY +
+        0.2 * opZ * dm.state * opZ ≈ apply(dm, ch).state
```
GiggleLiu, line 45: :+1: Tests look good. Maybe also cover the missed lines.

GiggleLiu (changes requested):
Thank you, it is a great contribution to Yao. It should be ready to go after fixing the issues.

lib/YaoArrayRegister/src/density_matrix.jl
jlbosse: Do we want to print `state` at all here? Or only have something like 
```
DensityMatrix{2, ComplexF64, Array...}
    active qubits: 3/3
    nlevel: 2
```
as the output, similar to `show(::ArrayReg)`?

lib/YaoBlocks/src/channel/error_channel.jl
jlbosse: The corresponding code in QuEST is found [here](https://github.com/QuEST-Kit/QuEST/blob/23a0a088dfd76d89f3252bffabda0387e880cda6/QuEST/src/CPU/QuEST_cpu.c#L399:551). It doesn't look like they have to apply 16 different Kraus operators and then sum the results up. But I also don't quite understand what they are doing instead.

lib/YaoArrayRegister/src/density_matrix.jl
jlbosse: I have some code for a function `partial_tr_outer(state::AbstractVector, outindices)`
that computes the partial trace of the outer product `state * state'` on `outindices` without constructing the outer product explicitly. I can use that instead. And should I also add it as an interface to `YaoAPI/src/registers.jl`?

lib/YaoArrayRegister/src/density_matrix.jl
jlbosse: It is among the lines not covered by the tests, so I suppose we don't really need this. Happy to remove it and only add it once it turns out we actually need it.

lib/YaoArrayRegister/src/density_matrix.jl
GiggleLiu: Maybe `density_matrix(rand_state(T, 2n; nlevel), n+1:2n)` is what you need ;D.

lib/YaoArrayRegister/src/density_matrix.jl
GiggleLiu: Both are good to me. It is not good to show every row, since it might flood users' terminal. For array register printing, I can not remember why we choose not printing the state. @Roger-luo

lib/YaoBlocks/src/channel/error_channel.jl
GiggleLiu: Thanks for the link, maybe we can leave a issue and resolve it later.

lib/YaoArrayRegister/src/density_matrix.jl
Roger-luo: For register we have the concept of active qubits, I think back then it was too verbose to print the full state, and we decide to just let user call `state` if they want to see the full state.

Roger-luo: I'm not sure if printing the full state even makes sense as default, it's too verbose and doesn't provide much information except for the simple cases.

GiggleLiu: We can use `print_matrix`, which prints `...` if the matrix size is too large.

lib/YaoArrayRegister/src/density_matrix.jl
jlbosse: Yeah, that makes sense. Though of course that would still mean we have type piracy that we eventually need to fix, right?

lib/YaoArrayRegister/src/density_matrix.jl
jlbosse: Ah, yes of course. Thanks!

lib/YaoArrayRegister/src/density_matrix.jl
jlbosse: I agree with @Roger-luo here and think that the printing of an `ArrayRegister` and a `DensityMatrix` should be as similar as possible. If the user actually wants to see the whole matrix they can still use `print_matrix(dm.state)`.

lib/YaoBlocks/src/channel/error_channel.jl
jlbosse: Opened an issue with the link

lib/YaoArrayRegister/src/density_matrix.jl
Roger-luo: We don't need to fix it here. But I want to clarify type piracy is not just about changing someone's code causing dispatch error now - it effects how code get cached, to Julia compiler this still looks like two packages, it does not know it belongs to the same developer, this no code caching will happen for any combination of these functions. It might be fine for now, but it's not scalable

lib/YaoArrayRegister/src/density_matrix.jl
Roger-luo: yeah, let us be consistent on printing, and this PR should do the same thing as `ArrayReg`, if we really decide to change the printing, we should do that in a separate PR and change both of them.

And regarding printing the full state, I hope it can be less verbose than the default matrix printing, the size it shows up in addition to qubits, levels, etc is still too many rows to me, and most of the corresponding entries do not mean anything.

Roger-luo: > We can use print_matrix, which prints ... if the matrix size is too large.

regarding `print_matrix`, it's not a stable API so it's not supposed to be called by an external package other than `Base`, if we depend on this API, we will be responsible for potential changes in new Julia versions,  it's better to just call `show(::IO, ::MIME"text/plain", x)` and strip out the title if we want then insert the right indent to the printing. This is at least an exposed API. 

> If the user actually wants to see the whole matrix they can still use print_matrix(dm.state).

also calling `dm.state` or `st.state` will show the vector in the same way already, no need to call that function.

jlbosse: I changed show(io, ::DensityMatrix) to match the behaviour of show(io, ::ArrayRegister) now. Pretty printing of the state can be a future PR

jlbosse:
I moved `DensityMatrix` back into `YaoAPI` and I _think_ all other comments were also addressed. Please let me know if there is anything else that needs to be done, before this can get merged.

GiggleLiu (approved):

Roger-luo:
@jlbosse great work! thanks!

PR #460, opened 2023-03-07
Added copyto!(::DensityMatrix)

jlbosse:
Same motivation as adding `similar(::DensityMatrix)`: It is desirable to get as many features of `ArrayReg` as possible to work for `DensityMatrix` as possible, so that most code written with a `ArrayRegister` in mind also works with  `DensityMatrix`.


jlbosse:
The CI failure seems to have something to do with julia servers being down, not with this PR

Roger-luo:
LGTM

PR #461, opened 2023-03-08
More density matrix support

jlbosse:
I added tests and docstrings to #408 . 

Before this gets merged, two more questions:

 1.  Currently `qubit_type()` is defined only for `YaoArrayRegister` and `DensityMatrix`. But its definition is sufficiently general, that it could be moved into `YaoAPI`. Shall I do that?
 2. More aesthetic: Currently a generic density matrix is referred to as `ρ`, `rho`, `r` or `dm` in the different files of the code (and also has different names in different docstrings). Should I go ahead and homogenize the naming there, and if yes to which of the four options?


Roger-luo:
> Currently qubit_type() is defined only for YaoArrayRegister and DensityMatrix. But its definition is sufficiently general, that it could be moved into YaoAPI. Shall I do that?

In a separate PR, yes.

> More aesthetic: Currently a generic density matrix is referred to as ρ, rho, r or dm in the different files of the code (and also has different names in different docstrings). Should I go ahead and homogenize the naming there, and if yes to which of the four options?

Yes, they should use the same variable name, but this is kinda written by different people at different times, so we haven't been careful. And since you are doing it, you have the right to pick your favorite except `r` (it's too short as a general convention)

But please also do this in a separate PR. I think this PR is good on its own already.

Roger-luo (approved):

GiggleLiu:
qubit_type() returns a string for printing. It is not an essential API that people will be interested in implementing it. We have to keep YaoAPI easy to understand and easy to extend, so any nonessential API should not go to YaoAPI.

PR #462, opened 2023-03-09
Moved `qubit_type` to `YaoAPI`

jlbosse:
The title says it all. Only open question from me is, whether we want to export `qubit_type` as part of the general API of an `AbstractRegister`

### DIFF
--- a/lib/YaoAPI/src/registers.jl
+++ b/lib/YaoAPI/src/registers.jl
@@ -29,6 +29,14 @@ For qubits, `D = 2`.
 """
 abstract type AbstractRegister{D} end
 
+"""
+    qubit_type(::AbstractRegister{2})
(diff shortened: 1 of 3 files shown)


Roger-luo:
after thinking about this a bit, and as @GiggleLiu [commented here](https://github.com/QuantumBFS/Yao.jl/pull/461#issuecomment-1461859758) this function is only used for printing not really address the basis information (it's inside the type parameters), and assuming someone is overloading it for say `rydberg` it's a bit strange that `qubit_type` is made for basis. This is a quite vague API IMO, we should eventually get rid of it with something more general. 

also, I don't think the basis is well supported in YaoBlocks, we probably shouldn't encourage people to put too much effort into building things on top of it.

jlbosse:
Fair enough, I will close this then

PR #469, opened 2023-05-15
Bumped compat for StatsBase

jlbosse:


GiggleLiu (approved):
Thanks for the version bump!

PR #476, opened 2023-07-11
Fixed KronBlock equality check

jlbosse:
Previously to this PR we had that 

```julia
using Yao
kb1 = kron(3, 1 => X)
kb2 = kron(3, 1=>X, 2=>Z)
kb3 = kron(3, 1=>X, 2=>Z, 3=>Y)
kb1 == kb2

# output 
false  # as one would expect

kb2 == kb3

# output
ERROR: DimensionMismatch: arrays could not be broadcast to a common size; got a dimension with lengths 2 and 3
```

This is fixed with this PR.


GiggleLiu:
The changes look good to me, thank you!

PR #479, opened 2023-09-25
Fix symbolic abs function

GiggleLiu:
This PR fixes the previous incorrect symbolic abs function. Depends on another PR to be merged: https://github.com/symengine/SymEngine.jl/pull/267

This PR also explicitly add the SymEngine as the dependency. I notice that two minor versions has been released, and it seems most stasbility issue in SymEngine has been resolved.


GiggleLiu:
@Roger-luo This PR is ready for review. It fixes some urgent bugs in YaoSym.

Roger-luo (approved):

PR #483, opened 2023-11-02
Changed TimeEvolution to use KrylovKit instead of ExponentialUtilities

jlbosse:
Implemented the changes discussed on Slack for more robust `TimeEvolution` gates.

I will also create a companion PR on CuYao that also removes the dependence on ExponentialUtilities


Roger-luo (approved):
that's a fast PR! Thank you

GiggleLiu:
Thanks, but I thought EU should be retained as an alternative backend. Any specific reason why it is removed while it is faster in some senarios?

jlbosse:
Ah sorry, yes I jumped the gun here. Rereading the discussion on Slack we did say to have both options and only make KrylovKit the default backend. I will try to create new PR this weekend doing that.

In terms of patch or breaking release, I suppose there is no way of "undoing" the current breaking release after we added EU support back in?

GiggleLiu:
> Ah sorry, yes I jumped the gun here. Rereading the discussion on Slack we did say to have both options and only make KrylovKit the default backend. I will try to create new PR this weekend doing that.
> 
> In terms of patch or breaking release, I suppose there is no way of "undoing" the current breaking release after we added EU support back in?

A new PR would be perfect. Thank you!

PR #488, opened 2023-12-07
transpsoe block

Roger-luo:
here is an example of how to implement transpose block, I haven't impl any tests here, so this is only for reference.


lib/YaoBlocks/src/composite/tag/transpose.jl
```diff
@@ -0,0 +1,58 @@
...
+julia> struct QFT <: PrimitiveBlock{2} n::Int end
+
+julia> YaoBlocks.nqudits(q::QFT) = q.n
+
+
+julia> circuit(q::QFT) = qft(nqubits(q));
+
+julia> YaoBlocks.mat(x::QFT) = mat(circuit(x));
+
+julia> QFT(2)'
```
GiggleLiu, line 38: This example should be fixed.

GiggleLiu (commented):
Looks great, however, tests are missing.

GiggleLiu:
Oops, this PR is not for review. Then no worries.

PR #493, opened 2024-02-20
Include YaoPlots as a component package

GiggleLiu:
# List of changes
1. Include YaoPlots as a component package.
image

The using time increased from 0.68s to 0.9s on my machine.
```julia-repl
julia> @time using Yao
  0.901292 seconds (1.28 M allocations: 86.938 MiB, 5.83% gc time, 22.79% compilation time: 94% of which was recompilation)
```

2. Fixed some of the documentation.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -39,6 +39,7 @@ jobs:
           - YaoArrayRegister
           - YaoBlocks
           - YaoSym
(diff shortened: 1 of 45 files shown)


GiggleLiu:
The autodiff breaks on nightly, which is due to an error in Zygote.

Roger-luo (approved):

PR #494, opened 2024-02-24
Make CuYao an extension

GiggleLiu:
Instead of installing CuYao, the CUDA extension can be access by typing
```julia
julia> using Yao, CUDA

julia> reg = rand_state(5) |> cu
ArrayReg{2, ComplexF64, CuArray...}
    active qubits: 5/5
    nlevel: 2
```

To test the CUDA extension, please run
```bash
$ make test-CuYao
```

Some extra make targets are added:
- `make init` and `make init-docs` to initialize `Yao` and `docs` environment.
- `make update` and `make update-docs` to update the above environments.

- `make test-<package name>` to test a component package.
- `make test` to test all component packages.
- `make coverage` to test all component packages and generate the coverage files.
- `make` is a combinaton of `make init` and `make test`

- `make-servedocs` to serve documents.
- `make clean` to clean up document build and `.cov` files.

Maybe we can remove `.ci` in the future.


GiggleLiu:
@Roger-luo all tests pass except the nightly.

Project.toml
```diff
@@ -13,8 +13,15 @@ YaoBlocks = "418bc28f-b43b-5e0b-a6e7-61bbc1a2c1df"
...
 
+[weakdeps]
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+
+[extensions]
+CuYao = "CUDA"
+
 [compat]
 BitBasis = "0.8, 0.9"
+CUDA = "4, 5"
```
Roger-luo, line 24: do we actually support both versions?

ext/CuYao/README.md
```diff
@@ -0,0 +1,86 @@
+**Build status**: [![][gitlab-img]][gitlab-url]
+
+[gitlab-img]: https://gitlab.com/JuliaGPU/CuYao.jl/badges/master/pipeline.svg
+[gitlab-url]: https://gitlab.com/JuliaGPU/CuYao.jl/pipelines
```
Roger-luo, line 4: this is not valid link anymore?

Project.toml
GiggleLiu: I think so? We can remove it later if its not compatible anymore.

ext/CuYao/README.md
GiggleLiu: I do not think so, let me remove it.

GiggleLiu:
> how is CUDA current being tested in CI? I don't see any test being setup in the CI?

I haven't setup the CI yet, we need to test locally. We do not have tests for CuYao for long time.

docs/Project.toml
```diff
@@ -7,6 +7,7 @@ FFTW = "7a1cc6ca-52ef-59f5-83cd-3a7055c09341"
 KrylovKit = "0b1a1467-8014-51b9-945f-bf0ae24f4b77"
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
 Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
+LiveServer = "16fef848-5104-11e9-1b77-fb7a48bbb589"
```
Roger-luo, line 10: do we actually need this to build the doc?

GiggleLiu: No, it is just convenient, otherwise you will have to start a temp environment and install LiveServer. Do you have a better option? Now the work flow is
```makefile
make init-docs
make servedocs
```

Roger-luo: I'd just use a separate `Project.toml` file for dev only within `.ci` directory. So you can keep the main environment clean. This is where I hate the stack env design... but LiveServer is not that big to download so prob fine if you insist.

ext/CuYao/README.md
Roger-luo: can we remove the link, I think this is good to merge, otherwise it's blocking the other PRs

GiggleLiu: I removed the README, and moved the contents to the documentation. Two new sections about CuYao and YaoPlots are added in the latest commit, addresing the issue.

docs/Project.toml
GiggleLiu: Oh I see. We can fix this issue in the future when this workflow causes any trouble.

PR #495, opened 2024-02-28
Fix tracedist 

exAClior:
Current calculation of trace distance is off by a factor of 1/2. Now the result matches wiki's and documentation's definition


Makefile
```diff
@@ -0,0 +1,37 @@
+JL = julia --project
+
+default: init test
```
GiggleLiu, line 3: Please do not commit this Makefile. It has already been included in another PR.

lib/YaoArrayRegister/src/density_matrix.jl
```diff
@@ -77,7 +77,7 @@ end
 YaoAPI.density_matrix(reg::ArrayReg{D}) where D = DensityMatrix{D}(reg.state * reg.state')
 YaoAPI.density_matrix(rho::DensityMatrix) = copy(rho)
 
-YaoAPI.tracedist(dm1::DensityMatrix{D}, dm2::DensityMatrix{D}) where {D} = trace_norm(dm1.state .- dm2.state)
+YaoAPI.tracedist(dm1::DensityMatrix{D}, dm2::DensityMatrix{D}) where {D} = trace_norm(dm1.state .- dm2.state) / 2.0
```
Roger-luo, line 80: can we also add a docstring about the definition?

Makefile
exAClior: sure I can remove it

lib/YaoArrayRegister/src/density_matrix.jl
exAClior: There is already a doc string [here](https://github.com/QuantumBFS/Yao.jl/blob/4c7f5cacef274faef3a6085aae8de291f8a51456/lib/YaoAPI/src/registers.jl#L655-L681). Should we still add a docstring here?

GiggleLiu:
LGTM. Thanks, @exAClior

PR #506, opened 2024-05-02
Fix CUDA extension - Project.toml

### LINKED ISSUE #505: CUDA is listed as a wrong compat
Why are we depending on CUDA here? I believe CUDA should be remove from compat and only appear in extension dependency @GiggleLiu 

https://github.com/QuantumBFS/Yao.jl/blame/ba48e22d15e0c97e41b46f4b1e0e039e657e9c46/Project.toml#L19

GiggleLiu:
fix #505 

In this PR,
- I moved the CUDA extension of YaoToEinsum to CuYao.
- Restored the `[weakdeps]` and `[extensions]` sections that removed in a previous commit.


Roger-luo:
Thanks

Issue #534, opened 2025-01-31
Potential Issue with "collapseto!" normalization for mixed states


FabianBallar7:
First of all, I’d like to thank everyone involved in adding and improving support for density matrices in Yao! These features are quite new, and it’s great to see them being developed.

This is my first time submitting an issue, so I’m happy to learn from any feedback. I apologize in advance if I made any mistake in reporting this, and I appreciate your time in reviewing it.

### Summary (Expected/Observed behavior)

I noticed that `collapseto!` sometimes results in an unnormalized density matrix
when collapsing onto a subspace. Specifically, when working with mixed states
(generated via `rand_density_matrix`), the trace after collapse is not consistently
1. 
 
### Steps to Reproduce
```julia
using Yao, LinearAlgebra
 test_s = rand_density_matrix(8) 
collapseto!(test_s, (5:8) => zeros(Int, 4)) # Collapse qubits 5-8 to |0000>
 println(tr(test_s.state)) # Expected: 1, but sometimes not
```

### Possible Explanation

The implementation of `collapseto!` for mixed states looks like this:

```julia
function YaoAPI.collapseto!(rho::DensityMatrix, locsval::Pair)
    locs = locsval.first isa AllLocs ? (1:nqudits(rho)) : locsval.first
    ic = itercontrol(nqudits(rho), collect(locs), locsval.second) .+ 1
    st = normalize!(rho.state[ic, ic])
    fill!(rho.state, 0)
    rho.state[ic, ic] .= st
    return rho
end
```

I believe the issue may be that when we look at the density matrix window this is no longer a DensityMatrix type and hence doesn't go through the `LinearAlgebra.normalize!(r::DensityMatrix)` function but rather the standard [`LinearAlgebra.normalize!`](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.normalize!) and most likely computing the Frobenius norm of the density matrix and not the trace.

For density matrices stemming from pure states, this shows no issue precisely because  $|A|_F = \sqrt{\text{Tr}\left(A A^{\dagger}\right)}$, and for a pure state $\rho^2 = \rho$ they yield the same value. 

### Potential Fix

It could be as simple as just dividing by the trace of the small window. Here's the function I implemented, which seems to work well. 

```julia
function collapse2!(rho::DensityMatrix, locsval::Pair)
    locs = locsval.first isa AllLocs ? (1:nqudits(rho)) : locsval.first
[...]

PR #535, opened 2025-01-31
Fix collapseto! normalization and add test for mixed states

FabianBallar7:
This PR fixes collapseto! to normalize density matrices properly, based on #534. Adds a warning when collapse results in a near-zero probability state. A test is added to ensure normalization remains valid in these cases.

### DIFF
--- a/lib/YaoArrayRegister/src/density_matrix.jl
+++ b/lib/YaoArrayRegister/src/density_matrix.jl
@@ -226,11 +226,10 @@ function YaoAPI.collapseto!(rho::DensityMatrix, locsval::Pair)
     st = rho.state[ic, ic]
     
     trace_norm = tr(st)  
-    if real(trace_norm) > 1e-10
-        st ./= trace_norm  
-    else
+    if real(trace_norm) <= 1e-10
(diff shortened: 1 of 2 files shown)


GiggleLiu:
The changes looks perfect, thank you!

lib/YaoArrayRegister/src/density_matrix.jl
```diff
@@ -223,7 +223,14 @@ end
...
     ic = itercontrol(nqudits(rho), collect(locs), locsval.second) .+ 1
-    st = normalize!(rho.state[ic, ic])
+    st = rho.state[ic, ic]
+    
+    trace_norm = tr(st)  
+    if real(trace_norm) > 1e-10
+        st ./= trace_norm  
+    else
+        @warn "Collapse resulted in a near-zero probability state. Normalization may be inaccurate."
+    end
```
GiggleLiu, line 233:
```suggestion
    if real(trace_norm) <= 1e-10
        @warn "Collapse resulted in a near-zero probability state. Normalization may be inaccurate."
    end
    st ./= trace_norm  
```

PR #538, opened 2025-02-18
Remove unnecessary const declarations when setting a ref

KristofferC:
These should probably have errored on earlier julia versions

Ref https://github.com/JuliaLang/julia/issues/57334#event-16251080756


Roger-luo:
ehhh... looks like CI is too old.

KristofferC:
Bump, if possible it would be nice with a new release for this to avoid some PkgEval noise.

Roger-luo:
Ok I think I'm gonna merge this branch first then fix the CI separately.

PR #566, opened 2025-08-17
Visualize YaoToEinsum

GiggleLiu:
## Updates
- Store the mapping from variable to qubits in the result.
- Using the result, calculate the coordinates of tensors for better visualization.
- Detect special tensors to interpret and annotate the tensors.
- Visualize with LuxorGraphPlot backend.

## Examples
1. Visualize a tensor network in vector mode (normal mode)
image

2. Visualize a tensor network in density matrix mode
image


GiggleLiu:
Since this is a simple PR about tensor network visualization, not much worth reviewing. I will merge it directly.

PR #582, opened 2026-03-30
Add a Power block to yaoblocks

harry-phasecraft:
Add a PowBlock. This is a block with a contents and a power that acts as $U^{pow}$ on the circuit.

This helps performance with large repetative circuits.


lib/YaoBlocks/test/autodiff/apply_back.jl
```diff
@@ -139,3 +139,15 @@ end
         )
     end
 end
+
+@testset "apply pow" begin
+    Random.seed!(5)
+    for reg0 in [rand_state(3), rand_state(3, nbatch = 10)]
+        for n in [-3,-2,-1,1,2,3]
```
jlbosse:
```suggestion
        for n in -3:3
```
to make sure we catch the edge case of `n==0`

lib/YaoBlocks/test/autodiff/mat_back.jl
```diff
@@ -150,3 +150,16 @@ end
         use_outeradj = use_outeradj,
     )
 end
+
+@testset "mat pow" begin
+    use_outeradj = false
+    for n in [-3, -1, 1, 2, 3]
```
jlbosse:
```suggestion
    for n in -3:3
```

lib/YaoBlocks/src/autodiff/apply_back.jl
```diff
@@ -135,6 +135,17 @@ function apply_back!(st, circuit::AbstractAdd, collector; in)
     (in, inδ)
 end
 
+function apply_back!(st, block::PowBlock{D, BT, PT}, collector) where {D, BT<:AbstractBlock, PT<:Integer}
+    blk = block.pow >= 0 ? content(block) : adjoint(content(block))
+    res = Any[]
```
jlbosse: Unlike `ChainBlock` we know what the content of `PowBlock` is. So can we maybe be more specific than `Any` here because we exactly know what will end up in `res`?

lib/YaoBlocks/src/autodiff/apply_back.jl
```diff
@@ -135,6 +135,17 @@ function apply_back!(st, circuit::AbstractAdd, collector; in)
     (in, inδ)
 end
 
+function apply_back!(st, block::PowBlock{D, BT, PT}, collector) where {D, BT<:AbstractBlock, PT<:Integer}
+    blk = block.pow >= 0 ? content(block) : adjoint(content(block))
```
jlbosse:
```suggestion
    iszero(block.pow) && return st
    blk = block.pow > 0 ? content(block) : adjoint(content(block))
```

lib/YaoBlocks/src/autodiff/mat_back.jl
```diff
@@ -104,6 +104,23 @@ function mat_back!(::Type{T}, rb::KronBlock, adjy, collector) where {T}
     return collector
 end
 
+function mat_back!(::Type{T}, rb::PowBlock{D, BT, PT}, adjy, collector) where {T, D, BT<:AbstractBlock, PT<:Integer}
+    nparameters(rb) == 0 && return collector
```
jlbosse, line 108:
```suggestion
    iszero(rb.pow) && return collector
    nparameters(rb) == 0 && return collector
```

lib/YaoBlocks/src/autodiff/mat_back.jl
```diff
@@ -104,6 +104,23 @@ function mat_back!(::Type{T}, rb::KronBlock, adjy, collector) where {T}
...
+    blk = content(rb)
+    adjy = rb.pow > 0 ? adjy : adjy'
```
jlbosse, line 110: I am not sure this is correct, AFAICT `adjy` is an external input that we accumulate matrix gradients on? So I don't think we need to conjugate here. Instead I believe we should do 
```
   blk = rb.pow > 0 ? content(rb) : content(rb)'
```
above.

lib/YaoBlocks/src/composite/pow_block.jl
```diff
@@ -0,0 +1,68 @@
+export PowBlock, power
+
+using LinearAlgebra
+
+"""
+    PowBlock{D,GT<:AbstractBlock} <: AbstractContainer{GT,D}
```
jlbosse:
```suggestion
    PowBlock{D,GT<:AbstractBlock} <: AbstractContainer{GT,D}
    PowBlock(block, pow) -> PowBlock
```

lib/YaoBlocks/src/composite/pow_block.jl
```diff
@@ -0,0 +1,68 @@
...
+    blk = pb.pow >= 0 ? pb.content : adjoint(pb.content)
+    for _ in 1:abs(pb.pow)
+        YaoAPI.unsafe_apply!(r, blk)
+    end
+    return r
+end
+
+function YaoAPI.unsafe_apply!(r::AbstractRegister, pb::PowBlock)
+    YaoAPI.unsafe_apply!(r, matblock(mat(pb)))
+end
```
jlbosse: Why is this needed? Doesn't the above implemementation fully cover `unsafe_apply!`?

lib/YaoBlocks/src/composite/pow_block.jl
```diff
@@ -0,0 +1,68 @@
...
+    return pb.pow != 0 ? nparameters(pb.content) : 0
+end
+Base.adjoint(pb::PowBlock) = PowBlock(adjoint(pb.content), pb.pow)
+Base.:(==)(a::PowBlock, b::PowBlock) = a.pow == b.pow && a.content == b.content
+Base.copy(pb::PowBlock) = PowBlock(pb.pow, pb.content)
+cache_key(pb::PowBlock) = (pb.pow, cache_key(pb.content))
+PropertyTrait(::PowBlock) = PreserveAll()
+
+"""
+    power(pow::Int, block::AbstractBlock) -> PowBlock
```
jlbosse:
```suggestion
    power(block::AbstractBlock, pow::Int) -> PowBlock
```
to be consistent with the actual behaviour :)

lib/YaoBlocks/src/composite/pow_block.jl
```diff
@@ -0,0 +1,68 @@
...
+
+### Examples
+
+```julia
+power(3, X)           # applies X three times
+power(Rx(0.5))(3)     # lazy form
+```
+"""
+power(block::AbstractBlock, pow::Real) = PowBlock(block, pow)
+power(block::AbstractBlock) = @λ(pow -> power(block, pow))
```
jlbosse: Not sure this is needed. This pattern is often used when `n` is the number of qubits and we want to be able to flexibly determine this at the point of need. 

What we might want is a @λ version that leaves the number of qubits `n` the curried parameter, not the power. E.g. we want the `PowBlock` version of this stuff:

https://github.com/QuantumBFS/Yao.jl/blob/1b4569495590739cb0f02a060b068fc6d9bf2c01/lib/YaoBlocks/src/composite/chain.jl#L60-L86

lib/YaoBlocks/test/composite/pow.jl
```diff
@@ -0,0 +1,107 @@
...
+
+# Multi-qubit circuits to use as PowBlock content
+circ2 = chain(2, put(1 => Rx(0.3)), put(2 => Rz(0.7)), cnot(2, 1, 2))
+circ3 = chain(3, put(1 => H), kron(3, 1 => Rx(0.2), 2 => Ry(0.4)), put(3 => T))
+
+@testset "power constructor" begin
+    pb = power(X, 3)
+    @test pb isa PowBlock
+    @test pb == PowBlock(X, 3)
+end
```
jlbosse: Add also a test with `^`?

jlbosse (changes requested):
Thanks a lot for this! 

I left some comments to make sure that edge cases work correctly for reverse mode AD and also to make the `power` constructor match the `chain` constructor as much as possible.

lib/YaoBlocks/src/autodiff/apply_back.jl
harry-phasecraft: This isnt actually needed as nparamers(blk) == 0 when pow=0 but im happy to add it just for clarity sake

lib/YaoBlocks/src/autodiff/mat_back.jl
harry-phasecraft: again this isnt needed as if pow==0 nparameters is defined as 0

lib/YaoBlocks/src/autodiff/mat_back.jl
harry-phasecraft: i wouldnt disagree except that i couldnt get that to pass the tests and this did. The implemtation for Daggered also does this step

lib/YaoBlocks/src/composite/pow_block.jl
harry-phasecraft: This is for non integer powers

lib/YaoBlocks/src/autodiff/apply_back.jl
jlbosse: Good point, maybe then just a comment pointing out that this isn't needed would suffice.

lib/YaoBlocks/src/autodiff/mat_back.jl
jlbosse: Again, then just comment pointing that out so the next person isn't confused :D

lib/YaoBlocks/src/composite/pow_block.jl
jlbosse: Oh, that would expect why you made the type of `pow` parameteric instead of just `Int`. So maybe add a comment just above this `# fall-back for non-integer powers`?

lib/YaoBlocks/src/autodiff/apply_back.jl
jlbosse: We just discussed this and concluded that e.g. `RepeatedBlock` does the same `Any`, so this should be fine.

lib/YaoBlocks/src/autodiff/mat_back.jl
jlbosse: @GiggleLiu or @Roger-luo can maybe one of you weigh in here? We are both a little bit confused as to how `mat_back!` exactly should behave.

lib/YaoBlocks/src/composite/pow_block.jl
jlbosse: After checking the `ScaleBlock` implementation I don't think the lambdified version is needed at all.

jlbosse:
@GiggleLiu or @Roger-luo:
Can we ask either of you to give this a review? From my side the following questions are still open:

 - Should this be named `PowBlock` or `Power`? None of the other algebraic blocks `Add` or `Scale` has `Block` in the name.
 - Which of  `unsafe_getindex` or `unsafe_getcol` are we missing here? They are implemented for e.g. the `Scale`.
 - Should we keep the `power` constructor, or make `^` the one canonical constructor for this?

Roger-luo:
> Should we keep the power constructor, or make ^ the one canonical constructor for this?

I would be leaning to keep `^` as the only constructor given you can already use the struct constructor itself explicitly.

> Should this be named PowBlock or Power? None of the other algebraic blocks Add or Scale has Block in the name.

Yes `Power` makes sense to me.

> Which of unsafe_getindex or unsafe_getcol are we missing here? They are implemented for e.g. the Scale.

I believe this was something @GiggleLiu added for coefficient propagation? As long as the main gate tests pass I'm ok with having it, but in case you need it, `unsafe_getindex` is the one preferred

jlbosse:
One more thing that seems to be missing is decomposition of this into native gates. If I do 
```julia
    rules = [
        Yao.Optimise.merge_pauli,
        Yao.Optimise.merge_scale,
        Yao.Optimise.eliminate_nested,
        Yao.Optimise.raise_chain,
        Yao.Optimise.combine_similar,
        Yao.Optimise.to_basictypes
    ]
    simplified_circuit = Yao.Optimise.simplify(circuit, rules=rules)
```
with some `circuit` containing a `Power` I get some error about `length` being missing.

lib/YaoBlocks/src/composite/power.jl
```diff
@@ -0,0 +1,50 @@
...
+
+Repeat the same block `content` `pow` times
+"""
+struct Power{D,BT<:AbstractBlock,PT<:Integer} <: AbstractContainer{BT,D}
+    content::BT
+    pow::PT
+end
+
+function Power(content::BT, pow::PT) where {D,BT<:AbstractBlock{D},PT}
+    if pow isa Int && pow < 0 && !isunitary(content)
```
jlbosse:
```suggestion
    if pow < 0 && !isunitary(content)
```
since we already disallow non-integer `pow` in `Power` blocks

lib/YaoBlocks/src/treeutils/optimise.jl
```diff
@@ -67,6 +67,11 @@ end
 function to_basictypes(block::Subroutine{D,<:Measure}) where {D}
     map_address(content(block), AddressInfo(nqudits(block), [block.locs...]))
 end
+function to_basictypes(block::Power)
+    @assert block.pow isa Integer
```
jlbosse, line 71: do we still need this @assert?

lib/YaoBlocks/src/measure_ops.jl
```diff
@@ -55,6 +55,16 @@ function eigenbasis(op::ChainBlock)
...
 
+function eigenbasis(op::Power)
+    b = op.pow >= 0 ? content(op) : adjoint(content(op))
+    E = chain(b.n)
+    blks = chain(b.n)
+    Ei, Vi = eigenbasis(b)
+    push!(E, fill(Ei, abs(op.pow))...)
+    push!(blks, fill(Vi, abs(op.pow))...)
+    return E, blks
+end
```
jlbosse, line 61: The eigenvectors of `U^n` are those of `U` and the eigenvalues of `U^n` are the eigenvalues of `U` raised to the n-th power. So I think this should simply be 

```julia
function eigenbasis(op::Power) 
    E, V = eigenbasis(content(op))
    return E ^ op.pow, V
end
```

where you need to check the `E ^ op.pow` part again. Currently it might return a `Power` block again, but we probably want a `matblock` with the eigenvalues (to the `pow`-th power) on the diagonal.

lib/YaoBlocks/src/composite/power.jl
```diff
@@ -0,0 +1,50 @@
...
+    blk = pb.pow >= 0 ? pb.content : adjoint(pb.content)
+    for _ in 1:abs(pb.pow)
+        YaoAPI.unsafe_apply!(r, blk)
+    end
+    return r
+end
+
+function nparameters(pb::Power)
+    return iszero(pb.pow) ? 0 : nparameters(pb.content)
+end
```
jlbosse: Here I am not sure of the `iszero` check is correct, or whether this should always just return `nparameters(content(pb))`. We'd have to check the parameter dispatching and reverse mode AD code then.

jlbosse (commented):
Main comment is regarding `eigenbasis` which probably needs a small fix, otherwise this looks good now!

lib/YaoBlocks/src/treeutils/optimise.jl
harry-phasecraft: nope

lib/YaoBlocks/src/composite/power.jl
harry-phasecraft: For me the problem is that if pow is 0 then there are no parameters regardless of what the content is which means. Is this because it will mess with how you might think `dispatch` etc will work if this returns 0 ?

jlbosse: We definitely want that `dispatch!(circuit, rand(nparameters(circuit))` always works. And it seems to me this might be easier to achieve (that is: with fewer lines of code) if we always return `nparameters(content(powblock))` here. This might also simplify the reverse mode AD logic, where we might be able to to avoid branching on `pow == 0` and instead just end up with some empty for loops.

And conceptually I guess the thing is that `block ^ 0` means "repeat block 0 times" and not "do nothing".

lib/YaoBlocks/src/measure_ops.jl
jlbosse: This is solved now.

lib/YaoBlocks/src/composite/power.jl
jlbosse: I made the suggested changes now. Will need https://github.com/QuantumBFS/LuxurySparse.jl/pull/82 though before it fully works.

jlbosse:
@GiggleLiu or @Roger-luo I believe this is now ready for review from one of you. From my point of view it is all good to merge now.

GiggleLiu:
Thank @jlbosse  and @harry-phasecraft , thank you for your contribution. The code quality looks good. I have a routine question: **do you have a concrete application that requires this gate type?**

Whenever we add a new composite block type, it complicates the design. So we expect the gain to be concrete enough. For example here, the autodiff behavior is different from traditional blocks, more chances to have bugs.

GiggleLiu:
I ran a focused review against `a4195c8` and found four correctness issues that should be addressed before merge:

1. **Matrix reverse AD assumes unitary content.** In `lib/YaoBlocks/src/autodiff/mat_back.jl:107-120`, the recurrence at line 117 cancels factors using unitarity, but positive powers permit non-unitary blocks. This reproduces with:

   ```julia
   test_mat_back(ComplexF64, Power(2.0 * Rx(0.3), 2), [2.0, 0.3]; δ=1e-6) # false
   ```

   The zero-power case is also incorrect: invertible content can produce nonzero gradients, while `Power(0.0 * Rx(0.3), 0)` throws `SingularException` in `mat_back`, even though the matrix is the constant identity. Please use the general derivative of `U^n` for positive non-unitary content and explicitly emit one zero per content parameter for `n == 0`.

2. **`Power` is missing the block indexing interface.** There are no `unsafe_getindex` or `unsafe_getcol` methods, so both `(X^2)[bit"0", bit"0"]` and `(X^2)[:, bit"0"]` throw `MethodError`. Every other relevant composite block implements these operations.

3. **`PropertyTrait(::Power) = PreserveAll()` is incorrect for zero powers.** For example, `(2.0 * X)^0` has the identity matrix, but `isunitary` and `isreflexive` return false; `(im * X)^0` is Hermitian, but `ishermitian` returns false. This also makes a subsequent negative power reject an actually unitary block. Zero needs a property special case, or `Power` needs an exact fallback rather than unconditional forwarding.

4. **The plotting implementation does not synchronize wire frontiers.** Drawing `Power(put(2, 1=>X), 5)` leaves `CircuitGrid.frontier == [1, 0]`; a following gate on wire 2 is placed inside the power box. The box also uses the full `address` range rather than the content occupied locations. Please advance/barrier every wire spanned by the box after drawing the content and add a sparse-occupancy regression test.

Release metadata also looks inconsistent: this changes YaoBlocks and YaoPlots while their versions remain at `0.14.4` and `0.9.6`, but the root Yao version is bumped to `0.9.4`.

Stable CI passes. The YaoBlocks nightly failure is in unrelated `ReflectGate` tests. Codecov still reports 87.38% patch coverage with 13 changed lines uncovered.

jlbosse:
>  do you have a concrete application that requires this gate type?
Our main use case was (optimizing) trotter circuits and Quantum Phase estimation circuits. They naturally have a structure where one sub-circuit gets taken to some power. If we then want to further do some circuit rewriting on that sub-circuit, it is much cheaper to just rewrite the content of a `Power` block once, rather than doing it `pow` times within a long `ChainBlock`.   

Regarding your review comments: Thanks for the thorough review, I hope they should all be addressed now.

Sorry for the delay on this...

PR #583, opened 2026-03-30
Fixed O(n^2) precompilation work for islocs_inbounds and islocs_conflict

jlbosse:
`@assert_locs_inbounds n locs "error message"` and `@assert_locs_safe` got enormously slow for large `locs` tuples. The culprit appeared to be that `TupleTools.sort` is a recursive merge sort on tuple types. `_merge(::NTuple{i}, ::NTuple{j}, ...)` requires a distinct LLVM specialisation for each (i,j) pair with `i+j ≤ N`, giving O(N²) new specialisations per first sort of length `N`.

This fixes these issues by not relying on TupleTools.sort for homogeneous tuples. Additionally,it speeds up `islocs_inbounds` from O(n log(n)) to O(n) on all input types and speeds up `islocs_conflict` from O(n log(n)) to O(n) for integer tuples.

For my concrete circuits (which were on several hundred qubits) this sped up things from running out of memory after half an hour to a few seconds to construct these circuits.

### DIFF
--- a/lib/YaoArrayRegister/Project.toml
+++ b/lib/YaoArrayRegister/Project.toml
@@ -1,6 +1,6 @@
 name = "YaoArrayRegister"
 uuid = "e600142f-9330-5003-8abb-0ebd767abc51"
-version = "0.9.13"
+version = "0.9.14"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/lib/YaoArrayRegister/src/error.jl
+++ b/lib/YaoArrayRegister/src/error.jl
@@ -27,6 +27,15 @@ end
 
 
 _sort(x::Vector; by = identity) = sort(x, by = by)
+# TupleTools.sort recurses on sub-tuples, generating O(N²) distinct _merge
+# specialisations for a length-N tuple.  collect+sort! compiles once per (N,T)
+# with O(N) LLVM instructions and zero recursive sub-specialisations.
+function _sort(x::NTuple{N,T}; by = identity) where {N,T}
+    v = collect(x)
+    sort!(v, by = by)
+    return return NTuple{N}(v)
+end
+# Heterogeneous Tuple (e.g. Tuple{Int, UnitRange{Int}}): always small in practice.
 _sort(x::Tuple; by = identity) = TupleTools.sort(x, by = by)
 
 # NOTE: this method assumes its input is not empty, it gets rid of errors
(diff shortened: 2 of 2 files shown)


jlbosse:
Not sure if the CI failures are related. They all are inside YaoBlocks related to reverse mode AD with Zygote and complain about in-place modification of BitVectors. The new `islocs_conflict` function does use BitVectors, but not anywhere that should be affected by differentiation and it also does not show up in the stack trace of the failures.

jlbosse:
~~Running `make test` in my local fork of Yao.jl also doesn't produce any failures.~~

Running `make test` did not actually use my forked versions. Running the `YaoBlocks` with this version of `YaoArrayRegister` shows that these changes do in fact break the YaoBlocks tests. Which then raises the question: How can we avoid the excessive pre-compilation while still remaining Zygote friendly?

GiggleLiu:
## CI Failure Analysis

The 3 YaoBlocks test failures are all the same root cause: the new `islocs_conflict(::NTuple{N, T<:Integer})` method uses a mutable `BitVector` (`falses` + `setindex!`), which Zygote cannot differentiate through.

The `PutBlock` constructor calls `@assert_locs_safe` → `islocs_conflict(locs)`, and the "zygote add" test differentiates through `PutBlock` construction, hitting:

```
Mutating arrays is not supported -- called setindex!(BitVector, ...)
```

### Suggested fix

Replace the `BitVector`-based implementation with `allunique`, which is O(n) via hashing, non-mutating, and Zygote-safe:

```julia
function islocs_conflict(locs::NTuple{N,T}) where {N, T<:Integer}
    return !allunique(locs)
end
```

This avoids coupling YaoArrayRegister to ChainRulesCore and is simpler than the current implementation.

jlbosse:
I just implemented your suggested fix and pushed.

While this PR fixes my most pressing problems, I am still observing size-dependent compilation workloads; That is new methods are compiled whenever `@assert_locs_safe` is called with different sizes of `locs`--which makes sense when working with tuples where each size of tuple is a different type--and the time for this compilation grows with the size of `locs`---which I would hope is avoidable?

In my concrete example compilation of `@assert_locs_safe` for a 2000 qubit circuit takes 14s, but constructing the same circuit a second time after compilation takes only 60ms. Doing exactly the same on a 36 qubit circuit with exactly the same structure compilation takes only 0.3s and constructing the same circuit a second time after compilation takes only 820μs.

If you want I can open another issue about that and try to get back to it when I have a bit more time?

GiggleLiu:
> I just implemented your suggested fix and pushed.
> 
> While this PR fixes my most pressing problems, I am still observing size-dependent compilation workloads; That is new methods are compiled whenever `@assert_locs_safe` is called with different sizes of `locs`--which makes sense when working with tuples where each size of tuple is a different type--and the time for this compilation grows with the size of `locs`---which I would hope is avoidable?
> 
> In my concrete example compilation of `@assert_locs_safe` for a 2000 qubit circuit takes 14s, but constructing the same circuit a second time after compilation takes only 60ms. Doing exactly the same on a 36 qubit circuit with exactly the same structure compilation takes only 0.3s and constructing the same circuit a second time after compilation takes only 820μs.
> 
> If you want I can open another issue about that and try to get back to it when I have a bit more time?

Yeah, just file a new issue please. It should be respected as long as it is a practical use case.

